PR #4, opened 2020-06-05
set up community build settings and bug fixes 

guoqing-noaa:


guoqing-noaa:
1. added FV3GFS_NCIO_Fortran_FLAGS for GNU compilers

2. Moved gsdcloud to src/,  make appropriate changes in CMakeLists.txt and a few bug fixes

3. pietc.f90 and pvqc_tables.f90
```
    use kinds, only: dp,dpc
--> use kinds, only: r_kind, dp, dpc
```
  This change is to avoid compiling error

`error #6580: Name in only-list does not exist or is not accessible.   [DP]`

Don't know the exact reason, but either (1) removing all #ifdef block in kinds.F90 or (2) use r_kind first can solve this problem. 

4. updated CMakeLists.txt for util/EnKF/arw/src/enspreproc_regional.fd

5. for community vesion, libsrc/ and fix/ are removed and installation of NCEPLIBS is required to build GSI.  ush/build.comgsi is updated accordingly.

PR #12, opened 2020-06-11
fix a bug in EnKF that can cause segfault in a diagnostic print 

jswhit:
in src/enkf/mpi_readobs, if the ensemble spread in observation space exceeds a value of 1000 a print statement is triggered that tries to print an array (mem_ob) that has already been deallocated.  This causes a segfault on Orion.  The fix is simply to not print mem_ob.  

Does not affect results, but protects against a possible segfault if ensemble spread exceeds the threshold value.


jswhit:
@RussTreadon-NOAA  - this should probably get included in the GFS v16 branch since this bug could lead to a crash.

MichaelLueken-NOAA:
Hi Jeff,

Since this is only a single commit, I can move forward with the change.  Please create a new GitHub issue detailing the issue that you corrected, then I can merge the pull request.

Thanks,

Mike

jswhit:
Closes issue #15

Issue #17, opened 2020-06-15
bug fixes for regional EnKF, GNU compilers; change build.comgsi; move gsdcloud to src/


guoqing-noaa:
1. added FV3GFS_NCIO_Fortran_FLAGS for GNU compilers

2. Moved gsdcloud to src/,  make appropriate changes in CMakeLists.txt and a few bug fixes

3. pietc.f90 and pvqc_tables.f90
```
    use kinds, only: dp,dpc
--> use kinds, only: r_kind, dp, dpc
```
  This change is to avoid compiling error

`error #6580: Name in only-list does not exist or is not accessible.   [DP]`

Don't know the exact reason, but either (1) removing all #ifdef block in kinds.F90 or (2) use r_kind first can solve this problem. 

4. updated util/EnKF/arw/src/enspreproc_regional.fd/CMakeLists.txt

5. bug fixes for GNU compilers and regional EnKF.

6. updated ush/build.comgsi for community users

PR #18, opened 2020-06-15
GitHub Issue NOAA-EMC/GSI#17.  Bug fixes for regional EnKF, GNU compilers; change build.comgsi; move gsdcloud to src.

guoqing-noaa:
1. added FV3GFS_NCIO_Fortran_FLAGS for GNU compilers

2. Moved gsdcloud to src/,  make appropriate changes in CMakeLists.txt and a few bug fixes

3. pietc.f90 and pvqc_tables.f90
```
    use kinds, only: dp,dpc
--> use kinds, only: r_kind, dp, dpc
```
  This change is to avoid compiling error

`error #6580: Name in only-list does not exist or is not accessible.   [DP]`

Don't know the exact reason, but either (1) removing all #ifdef block in kinds.F90 or (2) use r_kind first can solve this problem. 

4. updated util/EnKF/arw/src/enspreproc_regional.fd/CMakeLists.txt

5. bug fixes for GNU compilers and regional EnKF.

6. updated ush/build.comgsi for community users


hu5970:
Should we change directory name from GSD to NoVarCloudAnal for clarification.

guoqing-noaa:
> Should we change directory name from GSD to NoVarCloudAnal for clarification.

I would say better to leave it as is.  I had thought to change GSD to GSL. But keeping it as is will involve less changes and many community users are already familiar with this GSD naming convection.  May change it in the future if warranted. Thanks!

PR #20, opened 2020-06-15
Merge develop from feature/files_to_jedi_aop20 to feature/files_to_jedi

emilyhcliu:
This PR includes the following changes:
(1) updates modules for HERA
(2) adds and modifies variables for radiance, ozone, and SST to be written into diagnostic files
(3) add new GSI script (gen_diags_jedi_nrt.sh) to handle new data types


emilyhcliu:
I think I open the PR in the wrong place, I should open this PR in the forked branch.

Issue #28, opened 2020-07-07
regression test for aerosol impacts on radiance 


weiwilliam:
To secure the functionality of aerosol radiance impacts in GSI,

1. A new regression test, fv3aerorad, is added into the package.
    Two fix files, fv3aerorad_anavinfo and  fv3aerorad_satinfo.txt, are also included for the regression test.
    This regression test is based on fv3aero which is for aerosol analysis.
    
2. Bugfix for reading external aerosol files in read_files.f90.
     The capability is for GSI to read aerosol information from external files.
     Before the fix, GSI can't count "aerf" files properly.
     External reading for aerosol information is also utilized in "fv3aerorad" regression test.

MichaelLueken-NOAA:
PR #32 was merged to NOAA-EMC/GSI at 13366a1.  Closing issue.

PR #30, opened 2020-07-08
regression test for aerosol impacts on radiance

weiwilliam:
It's going to add a new regression test to secure the functionality of aerosol impact on radiance observations.

Please see issue #28 for more detail.


weiwilliam:
The version of submodule is not updated. Close it for modification.

PR #32, opened 2020-07-09
GitHub Issue NOAA-EMC/GSI#28.  regression test for aerosol impacts on radiance.

weiwilliam:
Add a regression test for aerosol impacts on radiance and bugfix for reading the external aerosol file.

Please see issue #28 for more detail


regression/regression_var.sh
```diff
@@ -122,7 +123,7 @@ case $machine in
 
    export fixcrtm="/scratch1/NCEPDEV/da/Michael.Lueken/CRTM_REL-2.2.3/crtm_v2.2.3/fix_update"
    export casesdir="/scratch1/NCEPDEV/da/Michael.Lueken/noscrub/CASES"
-   export ndate=$NDATE
+   export ndate=${NDATE:-/scratch2/NCEPDEV/nwprod/NCEPLIBS/utils/prod_util.v1.1.0/exec/ndate}
```
CoryMartin-NOAA, line 126: Is this necessary? Is prod_util not loaded by the regression tests?

regression/regression_var.sh
```diff
@@ -279,6 +283,7 @@ export gps_dtype="gps_bnd"
 export regression_vfydir="$noscrub/regression"
 
 # Define debug variable - If you want to run the debug tests, set this variable to .true.  Default is .false.
+#export debug=".true."
```
CoryMartin-NOAA, line 286: Should this commented line be added?

regression/global_C96_fv3aerorad.sh
```diff
@@ -0,0 +1,344 @@
...
+for loop in $loops; do
+
+case $loop in
+  01) string=ges;;
+  03) string=anl;;
+   *) string=$loop;;
+esac
+
+#  Collect diagnostic files for obs types (groups) below
+   listall="modis_aod_terra modis_aod_aqua"
```
CoryMartin-NOAA, line 318: Is this correct? I thought diag_aero was set to false in the namelist? Do we need to save diag files for other obs types?

weiwilliam: I also feel it's not needed section, because the script exit after the GSI finished, line 291.
Since this script is based on global_C96_fv3aero.sh, so I kept the section about diagnostic files.

regression/regression_var.sh
weiwilliam: I believe it can be removed.
I added it just for convenience to switch in longer wall-time.

CoryMartin-NOAA:
@weiwilliam you will also need to make changes to run the regression tests using CMake/ctest.

PR #63, opened 2020-09-16
Added a NC005091 (VIIRS winds in new BUFR) bug fix into feature/gfsda.v16.0.0_EUM_MetAVHRR_bufr

ilianagenkova:
When the new BUFR format for VIIRS winds was implemented in GSI  (in preparation for NOAA-20 VIIRS winds) a satellite ID range check was set incorrectly.
Satellite ID range [200 250] replaces the incorrect [250 299] range.

The other code changes allow for GSI to handle EUMETSAT and Metop/AVHRR(from NESDIS) winds in the new BUFR format adopted by all winds producers.

Routines with code changes:
read_obs.F90
read_satwnd.f90


src/gsi/read_satwnd.f90
```diff
@@ -1089,7 +1180,9 @@ subroutine read_satwnd(nread,ndata,nodata,infile,obstype,lunout,gstime,twind,sis
                 if(itype==251 )  then;  c_prvstg='GOESR' ; c_sprvstg='VIS'  ; endif
               endif
 ! Extra block for GOES-R winds: End
-
+           else ! wind is not recognised and itype is not assigned
+              write(6,*) 'READ_SATWND:SUBSET,itype, satID: ',trim(subset),itype,hdrdat(1)
```
RussTreadon-NOAA, line 1184: Is reaching this write statement a "normal' thing or do we want the write statement to be worded more strongly (e.g, include "***WARNING***"?

src/gsi/read_satwnd.f90
```diff
@@ -1089,7 +1180,9 @@ subroutine read_satwnd(nread,ndata,nodata,infile,obstype,lunout,gstime,twind,sis
                 if(itype==251 )  then;  c_prvstg='GOESR' ; c_sprvstg='VIS'  ; endif
               endif
 ! Extra block for GOES-R winds: End
-
+           else ! wind is not recognised and itype is not assigned
```
RussTreadon-NOAA, line 1183: This else block catches the situation when a subset read from satwnd does not have a corresponding subset section of code in read_satwnd.f90.  It does not catch the case in which a match is found but itype is NOT set for other reasons.   An example of this was subset NC005091.   itype remained -1 because the wrong satid range was checked.   itype-1 caused the code to fail when run with array bounds checking on.  If itype remains-1 after checking the subset, something is wrong, right?   If "yes", what do we want to happen in this situation?

RussTreadon-NOAA (commented):
Two comments in read_satwnd.f90

src/gsi/read_satwnd.f90
ilianagenkova: It is acceptable for a satwnd bufr file to contains winds that are not currently read if/when there are plans to be evaluated and possibly used. Such decision lead to keeping INSAT winds in the production satdumps. However, I also see the value of including "WARNING" -  it is a clear "inventory" of what is in the satwnd file, and what is used from him. I will add that.

src/gsi/read_satwnd.f90
ilianagenkova: I see your point. In that case I will add an additional:

if (itype==-1) then cycle loop_readsb   

after the two major if-then structures.

RussTreadon-NOAA (approved):
Code changes looks good.   Branch checked out on Venus and compiled twice - once with array bounds checking and a second time without array bounds checking.   Both executables run to completion.   No out of bounds arrays detected.  Bug fix results in VIIRS winds being assimilated as intended.   Third run made with uv 260 in monitor.  Analysis is similar but not bitwise identical with GFS v16 control.  The changes in this PR may be merged into release/gfsda.v16.0.0 pending GFS v16 management approval.

src/gsi/read_satwnd.f90
```diff
@@ -923,10 +949,79 @@ subroutine read_satwnd(nread,ndata,nodata,infile,obstype,lunout,gstime,twind,sis
...
+                 call ufbseq(lunin,amvqic,2,4,iret, 'AMVQIC') ! AMVQIC:: GNAPS PCCF
+                 qifn = amvqic(2,2)  ! QI w/ fcst does not exist in this BUFR
+                 ee = amvqic(2,4) ! NOTE: GOES-R's ee is in [m/s]
+              endif
+! Extra block for new Metop/AVHRR BUFR: End
+
 ! Extra block for VIIRS NOAA-20: Start
            else if(trim(subset) == 'NC005091') then
-              if(hdrdat(1) >=r250 .and. hdrdat(1) <=r299 ) then  ! The range of satellite IDs
+              if( hdrdat(1) >=r200 .and. hdrdat(1) <=r250  ) then  ! The range of satellite IDs
```
RussTreadon-NOAA, line 1024: Please revert this change as per management decision.   I recommend leaving the correct code in read_satwnd.f90 - just comment it out.   The corrected code will be activated in a later implementation.

RussTreadon-NOAA (commented):
Please update read_satwnd.f90 as indicated.

ilianagenkova:
read_satwnd.f90 has been updated to revert VIIRS winds fix

ilianagenkova:
Branch updated up to release/gfsda.v16.0.0

RussTreadon-NOAA (approved):
Changes look good.   Tests on Venus confirm that PR#63 reproduces v16rt2 control when setting uv 260 observation errors to the bufr missing value.   This PR may be merged into release/gfsda.v16.0.0

ilianagenkova:
Thank you, Russ!

Issue #76, opened 2020-10-29
Implement EFSOI process in FV3 GFS workflow


LiaofanLin-NOAA:
https://vlab.ncep.noaa.gov/redmine/issues/71386

PR #80, opened 2020-11-02
GitHub Issue NOAA-EMC/GSI#76 (insert GitHub issue number here).  Impl…

LiaofanLin-NOAA:
…ement EFSOI process in FV3 GFS workflow.


MichaelLueken-NOAA:
Review from Andrew Eichmann:

- Should scripts/fv3_save1.txt and scripts/fv3_save2.txt be in there? They look like saved copies of original scripts.

- in scripts/exglobal_enkf_update_fv3gfs.sh.ecf, for EFSOI PAOVERPB_THRESHOLD is set to 1.0 (turning off data discarding) - I vaguely remember that being discussed in this context, though I might be retconning the memory. Liaofan, could you remind me?

- in jobs/JGLOBAL_ENKF_UPDATE, there are suffixes set to nemsio - is this correct? Is this the EFSOI version that works with GFS v16? I thought there were more changes that were needed to handle NetCDF.

- for EFOSI purposes it looks right, my more general concern is that it doesn't interfere with GSI. Most of the changes are in the forecast sensitivity part, which I don't think anybody else uses, but I think some of the changes, particularly in src/enkf/params.f90, were effectively reversions from when I was trying to merge the old EFSOI code from some two years ago into the GSI of about a year ago, and some of it is probably at the very least superfluous. But mainly I don't want to break anything.

MichaelLueken-NOAA:
Liaofan's replies to Andrew's review:

> * Should scripts/fv3_save1.txt and scripts/fv3_save2.txt be in there? They look like saved copies of original scripts.

I am not sure about these two files.  These two were created when I merged the EXP-fsoi_fv3 branch with the master.

> * in scripts/exglobal_enkf_update_fv3gfs.sh.ecf, for EFSOI PAOVERPB_THRESHOLD is set to 1.0 (turning off data discarding) - I vaguely remember that being discussed in this context, though I might be retconning the memory. Liaofan, could you remind me?

This edit was likely originally from you.  You are right.  When PAOVERPB_THRESHOLD=1, we turn off data discarding.  However, since GFS v16 is using LETKF, the parameter might not be a matter. 

> * in jobs/JGLOBAL_ENKF_UPDATE, there are suffixes set to nemsio - is this correct? Is this the EFSOI version that works with GFS v16? I thought there were more changes that were needed to handle NetCDF.

Yes, the following two lines can be removed

export ASUFFIX=".nemsio"
export ESUFFIX=".ni.nemsio"

Though I found that $GSI/scripts/exglobal_enkf_update_fv3gfs.sh.ecf does not use ASUFFIX and ESUFFIX.  

> * for EFOSI purposes it looks right, my more general concern is that it doesn't interfere with GSI. Most of the changes are in the forecast sensitivity part, which I don't think anybody else uses, but I think some of the changes, particularly in src/enkf/params.f90, were effectively reversions from when I was trying to merge the old EFSOI code from some two years ago into the GSI of about a year ago, and some of it is probably at the very least superfluous. But mainly I don't want to break anything.

Yes, this is mainly for efsoi.x.  There is a need for another GSI-related merge(s) that work with the global workflow.  I will work on the global workflow merge later.

MichaelLueken-NOAA:
Andrew's follow-up:
 
> > * Should scripts/fv3_save1.txt and scripts/fv3_save2.txt be in there? They look like saved copies of original scripts.
> 
> I am not sure about these two files. These two were created when I merged the EXP-fsoi_fv3 branch with the master.
> 

I don't think I put them there - they can probably be removed.

> > * in scripts/exglobal_enkf_update_fv3gfs.sh.ecf, for EFSOI PAOVERPB_THRESHOLD is set to 1.0 (turning off data discarding) - I vaguely remember that being discussed in this context, though I might be retconning the memory. Liaofan, could you remind me?
> 
> This edit was likely originally from you. You are right. When PAOVERPB_THRESHOLD=1, we turn off data discarding. However, since GFS v16 is using LETKF, the parameter might not be a matter.
> 

I don't see where I would have gotten them from. I guess we can leave that there.

> > * in jobs/JGLOBAL_ENKF_UPDATE, there are suffixes set to nemsio - is this correct? Is this the EFSOI version that works with GFS v16? I thought there were more changes that were needed to handle NetCDF.
> 
> Yes, the following two lines can be removed
> 
> export ASUFFIX=".nemsio"
> export ESUFFIX=".ni.nemsio"
> 
> Though I found that $GSI/scripts/exglobal_enkf_update_fv3gfs.sh.ecf does not use ASUFFIX and ESUFFIX.
> 
> > * for EFOSI purposes it looks right, my more general concern is that it doesn't interfere with GSI. Most of the changes are in the forecast sensitivity part, which I don't think anybody else uses, but I think some of the changes, particularly in src/enkf/params.f90, were effectively reversions from when I was trying to merge the old EFSOI code from some two years ago into the GSI of about a year ago, and some of it is probably at the very least superfluous. But mainly I don't want to break anything.
> 
> Yes, this is mainly for efsoi.x. There is a need for another GSI-related merge(s) that work with the global workflow. I will work on the global workflow merge later.

Ok - I guess the switch to NetCDF needed less change than I expected.

scripts/exglobal_efsoi_fv3gfs.sh.ecf
```diff
@@ -0,0 +1,520 @@
...
+ls -ltr
+
+
+################################################################################
+#  Postprocessing
+cd $pwd
+
+#AFE
+#cp -r $DATA /scratch1/NCEPDEV/stmp4/Andrew.Eichmann/efsoi
+cp -ir $DATA /scratch2/BMC/rtrr/llin/200601_EFSOI_03/EFSOI_DUMP
```
aerorahul, line 509: Who owns this path?

jobs/JGLOBAL_EFSOI
```diff
@@ -0,0 +1,138 @@
...
+###############################################################
+# Run relevant exglobal script
+env
+msg="HAS BEGUN on `hostname`"
+postmsg "$jlogfile" "$msg"
+$LOGSCRIPT
+
+# LFL
+#SCRgsi=/scratch1/NCEPDEV/da/Andrew.Eichmann/fv3gfs/develop-efso/global-workflow/scripts/
+SCRgsi=/scratch2/BMC/rtrr/llin/200601_EFSOI_03/global-workflow_02/scripts/
```
aerorahul, line 99: Hardwired paths should not be allowed.

scripts/exglobal_efsoi_fv3gfs.sh.ecf
```diff
@@ -0,0 +1,520 @@
...
+################################################################################
+####  UNIX Script Documentation Block
+#                      .                                             .
+# Script name:         exglobal_efsoi_fv3gfs.sh.ecf
+# Modified from:         exglobal_enkf_update_fv3gfs.sh.ecf
+# Script description:  Make global_enkf update
+#
+# Author:        Rahul Mahajan      Org: NCEP/EMC     Date: 2017-03-02
+#
+# Abstract: This script runs the global_enkf update
```
aerorahul, line 11: If it does this, then how is it different from `exglobal_enkf_update.sh`

aerorahul (changes requested):
A quick review showed a lot of hard-wired paths that make it impossible for all users to use.

scripts/exglobal_efsoi_fv3gfs.sh.ecf
LiaofanLin-NOAA: Thank you for all the above comments.  I have updated exglobal_efsoi_fv3gfs.sh.ecf according.

jobs/JGLOBAL_EFSOI
LiaofanLin-NOAA: Thank you for the comments.  I have removed all the hardwired paths and updated JGLOBAL_EFSOI.

PR #82, opened 2020-11-06
fix an issue of reading nexrad bufr file

ShunLiu-NOAA:
FV3LAMDA sometimes can't read NEXRAD bufr files due to checking subset status before reading data. This checking step is not necessary and should be turned off.


MichaelLueken-NOAA:
These changes were merged to the NOAA-EMC/GSI master from https://github.com/NOAA-EMC/GSI/pull/83, so I will now close this extra PR.

Issue #81, opened 2020-11-06
fix the issue of reading NEXRAD bufr file


ShunLiu-NOAA:
FV3LAMDA sometimes can't read NEXRAD bufr files due to checking subset status before reading data. This checking step is not necessary and should be turned off.

Before the fix:
For today's 12z LAMDA:

regional_gsianl_tm01_12.log: RADAR_BUFR_READ_ALL:  problem opening level 2 bufr file "l2rwbufr"
regional_gsianl_tm02_12.log: RADAR_BUFR_READ_ALL:  problem opening level 2 bufr file "l2rwbufr"
regional_gsianl_tm03_12.log: RADAR_BUFR_READ_ALL:  problem opening level 2 bufr file "l2rwbufr"
regional_gsianl_tm04_12.log: RADAR_BUFR_READ_ALL:  problem opening level 2 bufr file "l2rwbufr"
regional_gsianl_tm05_12.log: RADAR_BUFR_READ_ALL:  problem opening level 2 bufr file "l2rwbufr"

File sizes for the 12z NAM nexrad file:

-rw-rw-r-- 1 nwprod prod 1218249536 Oct 21 13:15 nam.t12z.nexrad.tm00.bufr_d
-rw-rw-r-- 1 nwprod prod  829225392 Oct 21 12:18 nam.t12z.nexrad.tm01.bufr_d
-rw-rw-r-- 1 nwprod prod  951898336 Oct 21 11:23 nam.t12z.nexrad.tm02.bufr_d
-rw-rw-r-- 1 nwprod prod  951984000 Oct 21 10:30 nam.t12z.nexrad.tm03.bufr_d
-rw-rw-r-- 1 nwprod prod 1030995288 Oct 21 10:00 nam.t12z.nexrad.tm04.bufr_d
-rw-rw-r-- 1 nwprod prod 1072084336 Oct 21 10:00 nam.t12z.nexrad.tm05.bufr_d
-rw-rw-r-- 1 nwprod prod 1235644904 Oct 21 09:40 nam.t12z.nexrad.tm06.bufr_d

Conclusion: The smaller NAM nexrad files are the ones the code has problems reading. I recall seeing this before but this is the first time I was able to quantify it. It makes sense that the RAP files are giving the same error since they are always much  smaller than the NAM ones due to the early data cutoff.

after the bug fix, gsi can read NEXRAD bufr files from both RAP and NAM dump:

[...]

PR #83, opened 2020-11-06
GitHub Issue NOAA-EMC/GSI#81. Fix the issue of reading NEXRAD bufr file.

ShunLiu-NOAA:
fix the issue of GSI failure due to checking subset status before reading data. This checking step is not necessary and should be turned off.

### DIFF
--- a/src/gsi/read_l2bufr_mod.f90
+++ b/src/gsi/read_l2bufr_mod.f90
@@ -20,6 +20,7 @@ module read_l2bufr_mod
 !   2009-11-24  parrish  change time variable from regional_time (passed from gridmod) to
 !                          iadate (passed from obsmod), to prevent all radar data being tossed.
 !   2011-07-04  todling  - fixes to run either single or double precision
+!   2020-10-23  S.Liu    - fix reading l2rw_bufr issue
 !   2020-10-14  xu/sippel - Add options to superob the GRO
 !
 ! subroutines included:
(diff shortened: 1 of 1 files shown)


src/gsi/read_l2bufr_mod.f90
```diff
@@ -302,12 +303,6 @@ subroutine radar_bufr_read_all(npe,mype)
 !   Open bufr file with openbf to initialize bufr table, etc in bufrlib
     inbufr=10
     open(inbufr,file=infile,form='unformatted')
-    read(inbufr,iostat=iret)subset
-    if(iret/=0) then
```
TingLei-NOAA, line 306: After this fix,  the subset will not  be initialized. However, in the new line 338 block,  subset will be used.  Will this also be taken care of , like adding a  new logical branch?

ShunLiu-NOAA: Here, read(inbufr,iostat=iret)subset is not initialize "subset", it is to test if the file is readable. "subset" is declared as "character (8)", the first character is not necessary character(8). This caused the failure of iostat test. In line 338, "ireadmg" in bufr lib can handle this.

TingLei-NOAA: Got it , thanks. So, in the line 338,  do while(ireadmg(inbufr,subset,idate)>=0),  the subset  is to be read out rather than to be used as an intent-in argument.

TingLei-NOAA (approved):
The changes are ok to me.

MichaelLueken-NOAA:
The due date for the review committee has past with no feedback, so the changes have been merged to the master.

Issue #104, opened 2021-01-11
colliding filepaths


aerorahul:
The following files have the same name and pose a problem on a case insensitive filesystem e.g. macOS.

Please consider renaming one or the other with a distinct name.

```
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'util/Conventional_Monitor/image_gen/ush/Transfer.sh'
  'util/Conventional_Monitor/image_gen/ush/transfer.sh'
```

EdwardSafford-NOAA:
Fair enough.  I'll rename the lower case file name.

EdwardSafford-NOAA:
I've run this using the GFS/gdas ConMon image files and the transfer operation works fine.

MichaelLueken-NOAA:
Closing issue since the update has been merged to the master at 76aa328.

PR #105, opened 2021-01-13
Fix issue #104, remove file collision in ConMon

EdwardSafford-NOAA:
The ConMon file transfer scripts were overhauled and renamed using Transfer.sh and transfer.sh.  This created a potential file collision for MacOS and any other file systems that are case insensitive.  Renaming transfer.sh to transfer_imgs.sh fixes the problem.

### DIFF
--- a/util/Conventional_Monitor/image_gen/ush/Transfer.sh
+++ b/util/Conventional_Monitor/image_gen/ush/Transfer.sh
@@ -14,7 +14,8 @@ function usage {
 
 
 set -ax
-echo start Transfer.sh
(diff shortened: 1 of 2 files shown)


aerorahul (approved):
@EdwardSafford-NOAA 
I have not tested, but this looks correct.
Thanks for fixing this issue.

MichaelLueken-NOAA:
The changes have been approved by Rahul, so I will now commit these to the authoritative NOAA-EMC/GSI repository, since there are no changes to the source code.

PR #121, opened 2021-03-01
GitHub Issue NOAA-EMC/GSI #110.  Add hpc-stack to master.

RussTreadon-NOAA:
Changes are made to NOAA-EMC/GSI master `CMakeLists.txt`, `modulefiles.ProdGSI`, and `build_all_cmake.sh` to utilize hpc-stack (version `hpc/1.1.0`).  These changes are apply to WCOSS_D, Jet, and Orion.   hpc-stack was previously added to the master for Hera at b0a3bfd.

Issue #118, opened 2021-02-25
Implement EFSOI process in FV3 GFS workflow (2nd try)


AndrewEichmann-NOAA:
https://vlab.ncep.noaa.gov/redmine/issues/55974

AndrewEichmann-NOAA:
Modification of global_anavinfo.l127.txt  in fix committed and pushed in branch EXP-efso_fv3_PR_fix

PR #124, opened 2021-03-05
Implementation of EFSOI - EFSOI-specific fortran source files

AndrewEichmann-NOAA:
EFSOI-specific fortran source files and cmake file, all located in util/EFSOI_Utilities/src. This is the main code to make the EFSOI executable.

Files included:
CMakeLists.txt
efsoi.f90
efsoi_main.f90
gridinfo_efsoi.f90
gridio_efsoi.f90
loadbal_efsoi.f90
loc_advection.f90
scatter_chunks_efsoi.f90
statevec_efsoi.f90


AndrewEichmann-NOAA:
Associated with https://github.com/NOAA-EMC/GSI/issues/118

aerorahul:
@AndrewEichmann-NOAA 
Can you list what is changed in the `fix` and `libsrc` submodules?

util/EFSOI_Utilities/src/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+cmake_minimum_required(VERSION 2.6)
+if(BUILD_EFSOI)
+
+  file(GLOB LOCAL_SRC ${CMAKE_CURRENT_SOURCE_DIR}/*90)
+
+  message(echo ${PROJECT_BINARY_DIR})
```
aerorahul, line 6: Please remove debug statement.

util/EFSOI_Utilities/src/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+  
+  set_source_files_properties( ${LOCAL_SRC} PROPERTIES COMPILE_FLAGS ${ENKF_Fortran_FLAGS} )
+  
+  add_executable(global_efsoi.x  ${LOCAL_SRC} )
+  
+  set_target_properties( global_efsoi.x PROPERTIES COMPILE_FLAGS ${ENKF_Fortran_FLAGS} )
```
aerorahul, line 12: I don't think this is necessary.

util/EFSOI_Utilities/src/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+  
+  SET( CMAKE_EXE_LINKER_FLAGS  "${CMAKE_EXE_LINKER_FLAGS} ${OMPFLAG}" )
+  
+# copied from src/enkf/CMakeLists.txt   
+  include_directories(${CMAKE_CURRENT_BINARY_DIR} "${PROJECT_BINARY_DIR}/include/wrf" "${PROJECT_BINARY_DIR}/include/global"  ${CMAKE_CURRENT_BINARY_DIR}/.. ${MPI_Fortran_INCLUDE_DIRS} ${MPI_Fortran_INCLUDE_PATH} ${CORE_INCS} ${NETCDF_INCLUDE_DIRS} ${NCDIAG_INCS} ${FV3GFS_NCIO_INCS})
```
aerorahul, line 17: Why is the `wrf` directory included here?  There is no dependence on wrf in `efsoi` module or main program

util/EFSOI_Utilities/src/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+  
+  SET( CMAKE_EXE_LINKER_FLAGS  "${CMAKE_EXE_LINKER_FLAGS} ${OMPFLAG}" )
+  
+# copied from src/enkf/CMakeLists.txt   
+  include_directories(${CMAKE_CURRENT_BINARY_DIR} "${PROJECT_BINARY_DIR}/include/wrf" "${PROJECT_BINARY_DIR}/include/global"  ${CMAKE_CURRENT_BINARY_DIR}/.. ${MPI_Fortran_INCLUDE_DIRS} ${MPI_Fortran_INCLUDE_PATH} ${CORE_INCS} ${NETCDF_INCLUDE_DIRS} ${NCDIAG_INCS} ${FV3GFS_NCIO_INCS}) 
+
+#mostly copied from  src/enkf/CMakeLists.txt   
+  target_link_libraries( global_efsoi.x enkflib enkfdeplib ${GSILIB} ${GSISHAREDLIB} ${CORE_LIBRARIES}
+  ${MPI_Fortran_LIBRARIES} ${LAPACK_LIBRARIES} ${NETCDF_Fortran_LIBRARIES} ${NETCDF_C_LIBRARIES}
+  ${FV3GFS_NCIO_LIBRARIES} ${HDF5_Fortran_HL_LIBRARIES}
```
aerorahul, line 22: I think @RussTreadon-NOAA removed the explicit dependence on `HDF5`, `CURL` libraries.

util/EFSOI_Utilities/src/efsoi_main.f90
```diff
@@ -0,0 +1,166 @@
...
+ ! Needed for rad fixed files 
+ use radinfo, only: init_rad, init_rad_vars 
+ 
+ implicit none
+ integer :: ierr, nth
+ real(r_double) t1,t2
+
+ ! 0. initialize MPI.
+ call mpi_initialize()
+ if (nproc==0) call w3tagb('EFSOI_CALC',2018,0319,0055,'NP25')
```
aerorahul, line 83: The date is should be 2021, no?

util/EFSOI_Utilities/src/gridinfo_efsoi.f90
```diff
@@ -0,0 +1,461 @@
+module gridinfo_efsoi
```
aerorahul, line 1: Is this really necessary?
This is a big duplication of the code under `enkf`. We should be able to link to the `enkf` library for almost all of the needs that are needed.

util/EFSOI_Utilities/src/gridinfo_efsoi.f90
```diff
@@ -0,0 +1,461 @@
...
+         call stop2(24)
+      end if
+      call sigio_srhead(iunit,sighead,iret)
+      if (iret /= 0) then
+         print *,'error reading file in gridinfo ',trim(filename),' on task',nproc
+         call stop2(24)
+      end if
+      call sigio_sclose(iunit,iret)
+      ntrunc = sighead%jcap
+   endif
```
aerorahul, line 194: we are not using spectral model output for a while.  This should be cleaned.

util/EFSOI_Utilities/src/gridinfo_efsoi.f90
```diff
@@ -0,0 +1,461 @@
...
+          write(6,*)'grdinfo: gfs model: problem with nemsio_getfilehead, iret=',iret, ', file: ', trim(filename)
+          call stop2(23)
+       endif
+       print *,'ntrunc = ',ntrunc
+       if (nlons /= nlonsin .or. nlats /= nlatsin .or. nlevs /= nlevsin) then
+         print *,'incorrect dims in nemsio file'
+         print *,'expected',nlons,nlats,nlevs
+         print *,'got',nlonsin,nlatsin,nlevsin
+         call stop2(23)
+       end if
```
aerorahul, line 163: Do we need to keep capability for nemsio in FSOI computations?
@dtkleist 
If not, we should remove this option and simplify the duplications

util/EFSOI_Utilities/src/loadbal_efsoi.f90
```diff
@@ -0,0 +1,381 @@
+module loadbal_efsoi
```
aerorahul, line 1: how different is `loadbal_efsoi` from `loadbal` in the EnKF.
Again we should link to the `enkf` library and not try to duplicate code.

aerorahul (commented):
If you can address the comments first, we can move from there.

util/EFSOI_Utilities/src/gridinfo_efsoi.f90
dtkleist: I don't see why we would. Let's move forward with netcdf / v16 style.

util/EFSOI_Utilities/src/CMakeLists.txt
RussTreadon-NOAA: Yes, PR #121 removed explicit references to HDF5, CURL, and ZLIB.   Referencing NetCDF should be sufficient with hpc-stack.

AndrewEichmann-NOAA:
> @AndrewEichmann-NOAA
> Can you list what is changed in the `fix` and `libsrc` submodules?

It should be nothing, for now. I have an alternate global_anavinfo but that is in a directory under util that isn't in my master yet. Why does it come up?

aerorahul:
> > @AndrewEichmann-NOAA
> > Can you list what is changed in the `fix` and `libsrc` submodules?
> 
> It should be nothing, for now. I have an alternate global_anavinfo but that is in a directory under util that isn't in my master yet. Why does it come up?

When I click on "Files Changed", it shows changes in `fix` and `libsrc` submodules.

CoryMartin-NOAA:
@aerorahul could the changes in fix/ and libsrc/ be because `git submodule update` was not performed and these have been updated in the master since he first cloned?

util/EFSOI_Utilities/src/efsoi_main.f90
AndrewEichmann-NOAA: Should these be hardcoded at all?

https://www.nco.ncep.noaa.gov/pmb/docs/libs/w3lib/w3tagb.html

AndrewEichmann-NOAA:
Could I have swept them in with the merge using a recursive clone?

CoryMartin-NOAA:
I'm not a git expert by any means but if you merge in master and the fix/ directory hash is updated I think you have to then run git submodule update.

util/EFSOI_Utilities/src/CMakeLists.txt
AndrewEichmann-NOAA: removed

util/EFSOI_Utilities/src/CMakeLists.txt
AndrewEichmann-NOAA: removed

util/EFSOI_Utilities/src/CMakeLists.txt
AndrewEichmann-NOAA: removed

util/EFSOI_Utilities/src/CMakeLists.txt
AndrewEichmann-NOAA: removed

util/EFSOI_Utilities/src/gridinfo_efsoi.f90
AndrewEichmann-NOAA: I'll look into this. It appeared from comments and documentation that Dave intended to use as much of the enkf code as possible, so there might reasons for the EFSOI versions, and revisiting their necessity was on my todo list.

util/EFSOI_Utilities/src/gridinfo_efsoi.f90
AndrewEichmann-NOAA: works for me

util/EFSOI_Utilities/src/loadbal_efsoi.f90
AndrewEichmann-NOAA: I'll revisit this

util/EFSOI_Utilities/src/efsoi_main.f90
AndrewEichmann-NOAA: I guess everybody else hardcodes it

RussTreadon-NOAA: Good question.  What are the proper values to pass to `w3tagb`?   I agree that most developers probably cut-n-paste the `w3tagb `line from other codes with minimal changes.  This isn't right, but what's right?  We should ask EIB and/or NCO for their recommendation.

util/EFSOI_Utilities/src/gridinfo_efsoi.f90
aerorahul: Do we need this? What does it do other than print a header at execution (which is incorrect).
We have been copy/pasting stuff without really understanding the need for it for a while.

util/EFSOI_Utilities/src/efsoi_main.f90
aerorahul: Do we need this? What does it do other than print a header at execution (which is incorrect).
We have been copy/pasting stuff without really understanding the need for it for a while.

AndrewEichmann-NOAA: A scan of global-workflow shows all hardcoded, with dates going back to 1999. I could swear I've seen a line in a logfile that had the actual compilation date/time.

AndrewEichmann-NOAA: I would think it could be helpful if used correctly

MichaelLueken-NOAA:
@aerorahul @CoryMartin-NOAA @dtkleist @RussTreadon-NOAA - Should util/EFSOI_Utilities (and similarly, util/Correlated_Obs and util/*_Monitor) be moved from the util directory to src?  Changes in util are not reviewed and are usually treated as "use at your own risk."  With this being the case, the only change that would be necessary in this PR would be @aerorahul and @RussTreadon-NOAA comments for util/EFSOI_Utilities/src/CMakeLists.txt (for consistency).  Moving these from util to src would require additional review (like what is being down currently with these changes).  Should we move operational packages out of util and into a new home under src (we could add src/util to differentiate the two)?

Also, @CoryMartin-NOAA is correct with respect to both the fix and libsrc submodules.  Merging the NOAA-EMC/GSI master into your fork will update the submodules, but if you don't use git submodule update afterwards, the updated fix and libsrc won't be merged into your branch.

aerorahul:
Every piece of code that gets added to this (any) repository should be reviewed, IMO.

We don't do that for the GSI for whatever reason.
There are utilities in `util` that are used in operations `util/EnKF/gfs/src` and the monitoriing tools.  Are these also not reviewed?  There are `scripts` that are used in operations.  Are we not reviewing these either (atleast internally)? 
A later PR will add EFSOI scripts.

Moving these utilities from `util/` to `src/` does what exactly?  Why can they not be reviewed while being in `util/`, which is what they are.

`src/` is (was/should be) the directory for GSI and EnKF source codes.  Lately libraries `fv3gfs_ncio`, `ncdiag` etc. have been added to this directory.  Shouldn't these be part of `libsrc` (if at all)? If not, then in a dependency directory e.g. `deps` at the same level as `src`.

Code (Fortran/shell/python) etc. added should be reviewed.  If this is not possible within GSI, the utilities should be moved out to another targeted repository for utilities. 

I would like (all) code to be reviewed, when being added to `develop` is a long way of saying this.

util/EFSOI_Utilities/src/gridinfo_efsoi.f90
AndrewEichmann-NOAA: I've removed gridinfo_efsoi.f90, now using regular enkf version

util/EFSOI_Utilities/src/gridinfo_efsoi.f90
AndrewEichmann-NOAA: gridinfo_efsoi.f90 removed, using enkf version

util/EFSOI_Utilities/src/gridinfo_efsoi.f90
AndrewEichmann-NOAA: removed

util/EFSOI_Utilities/src/efsoi_main.f90
AndrewEichmann-NOAA: This is the output:
     PROGRAM EFSOI_CALC HAS BEGUN. COMPILED 2021319.55     ORG: NP25
     STARTING DATE-TIME  MAR 16,2021  16:43:52.635   75  TUE   2459290

util/EFSOI_Utilities/src/CMakeLists.txt
AndrewEichmann-NOAA: library references to HDF5 and CURL removed

AndrewEichmann-NOAA:
I ran git submodule update on a fresh clone of my master:

(base) -bash-4.2$ git submodule update
(base) -bash-4.2$ git push
Everything up-to-date
(base) -bash-4.2$ 

It didn't appear to change anything.

aerorahul:
> I ran git submodule update on a fresh clone of my master:
> 
> 
> 
> (base) -bash-4.2$ git submodule update
> 
> (base) -bash-4.2$ git push
> 
> Everything up-to-date
> 
> (base) -bash-4.2$ 
> 
> 
> 
> It didn't appear to change anything. Should that be done on a recursive clone?

`git submodule update --init --recursive `

MichaelLueken-NOAA:
> I ran git submodule update on a fresh clone of my master:
> 
> (base) -bash-4.2$ git submodule update
> (base) -bash-4.2$ git push
> Everything up-to-date
> (base) -bash-4.2$
> 
> It didn't appear to change anything. Should that be done on a recursive clone?

@AndrewEichmann-NOAA

To correct fix:

cd fix
git checkout rev2
cd ..

To correct libsrc:

cd libsrc
git checkout master
cd ..

git add fix libsrc

git commit --amend

git push origin master --force

AndrewEichmann-NOAA:
Ok, done. Does that look the way it should?

AndrewEichmann-NOAA:
I just updated my master with changes to remove gridinfo_efsoi.f90, will look into loadbal_efsoi

util/EFSOI_Utilities/src/loadbal_efsoi.f90
AndrewEichmann-NOAA: The problem in part is that the native loadbal is expecting to work with statevec, which I don't immediately see how I can convince to do what I need (read in two forecasts and an analysis). I have an idea or two, which would involve copying the state vector to new variables, but that would involve a vigorous refactoring. I understand the maintenance concerns and general annoyance of duplicated code, but loadbal also doesn't seem to have drifted much since 2016.

AndrewEichmann-NOAA:
I noticed that there are references to nemsio in the existing enkf gridio - is the plan to take those out too?

util/EFSOI_Utilities/src/loadbal_efsoi.f90
AndrewEichmann-NOAA: @aerorahul Any thoughts about this?

Issue #100, opened 2020-12-18
RadMon Maintenance Items


EdwardSafford-NOAA:
Several small issues with the RadMon have cropped up.  None are major or significant so they will be combined into this ticket.  Items include:

- Update Copy operation to avoid copying only partial data sets.
- Reorganize gdas and common directories to remove the outdated version numbering scheme.
- Implement RadMon code changes requested changes from NCO.

EdwardSafford-NOAA:
Have updated the data extraction and copy functions.  Included in this was a change to the low observational counts mechanism.  The previous method flagged as "low" an obs count that was < avg - 2*sdv for a given instrument/channel.  That produced a lot of warnings especially for the cris_* instruments.  I changed the definition of "low" to < 67% of average and will let this run over the weekend to see if that works out better.  First tests are promising.

EdwardSafford-NOAA:
Have updated the global angle plots.  Found that the current master version creates some empty channel plots.  That behavior has been corrected.  Was able to reduce job submission times and memory.

EdwardSafford-NOAA:
Committed first phase of updates for regional source.   Have DE, IG, and Xfer working for nam.  Have updated nam web site to reflect current sat/instruments.  Have begun generalizing clean_tankdir.sh to work with both glb and rgn sources.

EdwardSafford-NOAA:
Moved to hera to test there and found the executables no longer build using the package level (RadMon) cmake build script.  Per Mike the build scripts have to change to support hpc-stack libraries on hera (and I guess soon to be everywhere).  Things were updated to allow the GSI, enkf, etc to build on hera with those new libraries.  I figured out what needed to change in my RadMon script and eventually got it working.  I'll need to do that to the OznMon and ConMon but I'll open a separate ticket for those.

EdwardSafford-NOAA:
Hit a problem on hera on 2/26.  I had deleted most of my test data to reduce space.  When trying to move new test data for the RadMon using my new RSA SecureID app my permission is denied.  I tried the 8 digit numbers with and without my PIN, neither worked.  Opened a ticket with the hera help desk, situation is unchanged 3/1 AM.

EdwardSafford-NOAA:
Completed testing on hera, wcoss_c, wcoss_d.  Will wait for pr 110 (hpc-stack libs into cmake builds) before issuing a PR for this.

EdwardSafford-NOAA:
[...]

PR #127, opened 2021-03-10
PR RadMon maintenance #100

EdwardSafford-NOAA:
This PR completes issue #100, RadMon maintenance.


MichaelLueken-NOAA:
The changes were tested to ensure that the updated RadMon package compiles (due to updated CMakeLists.txt file).  RadMon compiles without issue in both standard and debug.  No changes were made to the source code, so these changes don't affect results. Giving final approval to this fix, then merging to the NOAA-EMC/GSI master.

Issue #128, opened 2021-03-11
Namelist options for commercial RO


KristenBathmann-NOAA:
Two namelist options will be added for commercial RO assimilation. These will be included in setup. The option gpstopalt will cutoff commercial data above this height. For example, setting gpstopalt=45.0 rejects all commercial data above 45 km. The option gpserrinf inflates commercial data errors. For example gpserrinf=1.5 multiplies the error by 1.5 at all levels. 

This work is being done under  [release/gfsda.v16.1.0](https://github.com/NOAA-EMC/GSI/tree/release/gfsda.v16.1.0)

dtkleist:
Do we want to make the names of these parameters more explicit so it is obvious that they apply to commercial GPSRO only and not all bending angle obs?

KristenBathmann-NOAA:
Sure, how about commgpstop and commgpserrinf?

dtkleist:
Works for me.

KristenBathmann-NOAA:
The new namelist options are working correctly. The cost function is lowered by cutting off commerical RO at 45 km, and by inflating the errors:
<img width="400" alt="cost4" src="https://user-images.githubusercontent.com/45015626/111173152-cd325980-857c-11eb-99d5-b34d263a5cfe.png">

Setting commgpstop to 55. (the current value for gpstop) allows all commerical RO below 55 km to be assimilated:

<img width="365" alt="spirectl" src="https://user-images.githubusercontent.com/45015626/111173288-e9ce9180-857c-11eb-908a-d0e2b555ccfa.png">

Setting commgpstop to 45. cuts off the commerical RO at 45 km:

<img width="365" alt="spirecut" src="https://user-images.githubusercontent.com/45015626/111173223-da4f4880-857c-11eb-84b4-f09c927b3c24.png">

Setting commgpstop to 45. does not affect other instruments:

<img width="365" alt="kompsatcut" src="https://user-images.githubusercontent.com/45015626/111173232-dd4a3900-857c-11eb-888b-d25efc552923.png">

MichaelLueken-NOAA:
These changes have been merged to the authoritative repo at 51a6564. Closing this issue.

PR #131, opened 2021-03-15
Commercial RO changes for Release/gfsda.v16.1.0

KristenBathmann-NOAA:
This PR adds namelist options associated with the assimilation of commercial RO. An update to the fix directory will be necessary to actually turn on the commercial data. See [issue #128](https://github.com/NOAA-EMC/GSI/issues/128)


CatherineThomas-NOAA (approved):
These code changes add two options for the assimilation of the commercial GNNSRO data.  The default is to act as the existing QC (commgpstop=gpstop and commgpserrinf=1).  These changes are straight forward and look good to me.

Issue #174, opened 2021-06-16
BUG:  append_diag missing from nc_diag_init calls


RussTreadon-NOAA:
Examination of `src/gsi/setupq.f90` identified a bug in subroutine `init_netcdf_diag_`. Logic in the routine checks to see if the netcdf  diagnostic file already exists. If it does, logical flag `append_diag` is set to `.true.`. This optional argument is passed to subroutine `nc_diag_init` to correctly set the file pointer. The `nc_diag_init` call in `setupq` does not include the optional argument `append_diag`.

GSI namelist `OBS_INPUT` for GFS v16 lists three sources of q observations

`prepbufr q null q 0.0 0 0`
`prepbufr_profl q null q 0.0 0 0`
`hdobbufr q null q 0.0 0 0`

When processing data from multiple bufr files it is necessary to set `append_diag` to `.true.` on subsequent calls to `setupq`.  Failure to do so means not all information for q observations is written to the netcdf q diagnostic file.   

The conventional monitoring package (ConMon) reads the netcdf q diagnostic file.   The absence of some q information in the netcdf q diagnostic file means ConMon can not accurately monitor q observations.   The bug reported in this issue likely explains the problem documented in issue #172.

The q observations assimilated by the LETKF are read from the netcdf q diagnostic file.   The absence of some q information in the netcdf q diagnostic file means the LETKF is not assimilating a complete set of q observations.   This impacts operational GFS v16 analysis results.   

A check of other setup routines in `src/gsi` finds the same bug in `setupcldtot.F90`, `setuplwcp.f90`, and `setupswcp.f90`.

This issue is opened to document this bug and its resolution.

RussTreadon-NOAA:
Add missing argument `append_diag` to `nc_diag_init` call in `setupcldtot.F90`, `setuplwcp.f90`, `setupq.f90`, and `setupswcp.f90`.   This commit, 2e7dfb0, was made to [feature/append_diag_bugfix](https://github.com/RussTreadon-NOAA/GSI/tree/feature/append_diag_bugfix) in [RussTreadon-NOAA /GSI](https://github.com/RussTreadon-NOAA/GSI) .

RussTreadon-NOAA:
**eobs, ediag, eupd test**

Use the 2021061500 cycle from the TAC2BUFR parallel, v16tac2bfr, to confirm that the changes committed at 2e7dfb0 fix the setupq problem.   

[...]

PR #176, opened 2021-06-16
GitHub Issue NOAA-EMC/GSI#174 correct append_diag bug

RussTreadon-NOAA:
As documented in NOAA-EMC/GSI issue #174, four setup routines in `src/gsi` are missing the optional argument `append_diag` in the call to `nc_diag_init`.   For those cases in which the GSI reads data for a given observation type from multiple input files variable `append_diag` is set to `.true.` so that the file pointer is correctly on subsequent calls to the given setup routine.   Without `append_diag` present, the resulting netcdf diagnostic files do not properly capture all information for the given observation type.   

This bug is present in the following `src/gsi` files:  `setupcldtot.F90`, `setuplwcp.f90`, `setupq.f90`, `setupswcp.f90`.

This bug impacts conventional data monitoring of q observations.   This bug also impacts the assimilation of q observations by the EnKF update code.

### DIFF
--- a/src/gsi/setupcldtot.F90
+++ b/src/gsi/setupcldtot.F90
@@ -922,7 +922,7 @@ subroutine setupcldtot(obsLL,odiagLL,lunin,mype,bwork,awork,nele,nobs,is,conv_di
         endif
      end if
 
-     call nc_diag_init(diag_conv_file)
+     call nc_diag_init(diag_conv_file, append=append_diag)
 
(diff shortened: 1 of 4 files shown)


EdwardSafford-NOAA (approved):

XiujuanSu-NOAA (approved):

MichaelLueken-NOAA:
The due date for the review committee has past with no feedback, so I will give final approval to these changes and merge them to the authoritative master.

Issue #180, opened 2021-06-28
 BUG: exgdas_enkf_select_obs.sh typo


RussTreadon-NOAA:
Lin Gan noticed the following error message in the job log file for enkf_select_obs

`chgrp: cannot dereference ‘obs_input.tar’: No such file or directory`

Investigation of the job found a typo in script `exgdas_enkf_select_obs`.    The following line 
`$NLN $SELECT_OBS obs_input.tar`

should read
`$NLN $SELECT_OBS obsinput.tar`

There is no `_` between `obs` and `input`.   This issue is opened to document correction of this typo.

RussTreadon-NOAA:
Create branch [feature/eobs_bugfix](https://github.com/RussTreadon-NOAA/GSI/tree/feature/eobs_bugfix) in [RussTreadon-NOAA/GSI](https://github.com/RussTreadon-NOAA/GSI) in which to correct the typo in `scripts/exgdas_enkf_select_obs.sh`.

RussTreadon-NOAA:
Test updated `exgdas_enkf_select_obs.sh` on Orion for 2020090306 case.   

The output below is from gdaseobs.log before the bug fix

```
634.999 + /bin/ln -sf /work/noaa/stmp/rtreadon/rocoto/prtest/gfs/prod/enkfgdas.20200903/06/atmos/gdas.t06z.obsinput.ensmean obsinput.tar
635.003 + chgrp rstprod obs_input.0001 obs_input.0002 obs_input.0003 obs_input.0005 obs_input.0009 obs_input.0010 obs_input.0011 obs_input.0017 obs_input.0018 obs_input.0029 obs_input.0030 obs_input.0031 obs_input.0041 obs_input.0055 obs_input.0057 obs_input.0060 obs_input.0061 obs_input.0062 obs_input.0063 obs_input.0066 obs_input.0068 obs_input.0069 obs_input.0070 obs_input.0072 obs_input.0073 obs_input.0075 obs_input.0076 obs_input.0087 obs_input.0088 obs_input.0089 obs_input.0090 obs_input.0098 obs_input.0099 obs_input.0100 obs_input.0101 obs_input.common obs_input.tar
chgrp: cannot dereference ‘obs_input.tar’: No such file or directory
[...]

PR #181, opened 2021-06-28
GitHub Issue NOAA-EMC/GSI#180.  correct typo in exgdas_enkf_select_obs.sh

RussTreadon-NOAA:
Issue #180 documents a bug found during TAC2BFR parallel testing.  Script `exgdas_enkf_select_obs.sh` contains a typo.   The filename `obs_input.tar` should be `obsinput.tar`.   The bug generates a `No such file` error message when the script is executed.  The bug does not alter analysis results.   

This PR is opened to merge the bug fix (correct the typo) into the authoritative master.

### DIFF
--- a/scripts/exgdas_enkf_select_obs.sh
+++ b/scripts/exgdas_enkf_select_obs.sh
@@ -72,7 +72,8 @@ cd $DATA || exit 8
 ################################################################################
 # ObsInput file from ensemble mean
 rm -f obs_input*
-$NLN $SELECT_OBS obs_input.tar
+rm -f obsinput.tar
+$NLN $SELECT_OBS obsinput.tar
 
 # Whether to save or skip obs
 if [ $RUN_SELECT = "YES" -a $USE_SELECT = "NO" ]; then


scripts/exgdas_enkf_select_obs.sh
```diff
@@ -72,7 +72,7 @@ cd $DATA || exit 8
 ################################################################################
 # ObsInput file from ensemble mean
 rm -f obs_input*
```
CatherineThomas-NOAA, line 74: @RussTreadon-NOAA 
Should there also be a "rm -f obsinput.tar" here? The exglobal_atmos_analysis.sh script has both "rm obs_input.*" and "rm obsinput.tar".

RussTreadon-NOAA: `rm -f obsinput.tar` has been added.  Good suggestion.

RussTreadon-NOAA:
Thank you, Cathy, for the prompt review.  

One potential problem with this PR.   This is a PR from my forked branch and not my forked master into the authoritative master.  I may need to abandon this PR and start over.   I'll wait for Mike's guidance.

CatherineThomas-NOAA:
@RussTreadon-NOAA At this point, I've only commented rather than given an official review, but since you've addressed it already, I can give my approval.  I'll also wait on Mike's guidance so I know which PR to approve.

RussTreadon-NOAA:
@CatherineThomas-NOAA    Yes, you are right.  We are only at the comment
stage of the PR.  The change has not been approved.   Two PR's have been
created.  I'll keep one and drop the other as per Mike's guidance.  If you
have additional comments, feel free to add them.  I'll work through them in
the final PR.

RussTreadon-NOAA:
> 
> 
> Thanks @RussTreadon-NOAA for addressing my comment on the rm line. I approve.

Thank you, Cathy.

MichaelLueken-NOAA:
Given @CatherineThomas-NOAA  approval and that the changes don't affect regression test results (as expected), I will now give final approval to these changes and merge them to the authoritative master.

Issue #175, opened 2021-06-16
Use the global 127L B-Matrix in regional FV3 DA (FV3LAMDA)


xyzemc:
@JacobCarley-NOAA 
FV3LAM DA currently still uses the old NAM B-Matrix, which needs to upgrade to FV3 model-based B-Matrix. Current GSI code m_berror_stats_reg.f90 is not able to directly read in the global 127L B-matrix. So GSI code needs to be changed for both reading the 127L B and correctly handling the B for regional FV3.

jderber-NOAA:
OK.  I will try to fix.

jderber-NOAA:
After several false paths, it was found that to use the global statistics in the regional model, there are 3 basic issues.

1. The global file is missing the second record that has the latitudes and sigmas of the statistics.
2. Some of the  arrays are dimensioned a bit differently in the global and regional files.  In the global, all arrays go from 1 to mlat.  In some of the regional arrays, they go from 0 to mlat+1.  This is not true of all arrays in the regional.
3. The global contains some additional variables (sst, ozone and cw) that are not available in the regional file.

To address these issues several changes were made.

1. A global file containing the latitudes and sigmas was constructed.  A namelist variable was included to allow it the user to supply the name of the input statistics file.  The default was not changed.  Also, a new namelist variable, usenewgfsberror logical namelist variable was included to allow the file with the extra record to be used in the global.  The same namelist variable was used for the regional to specify that the new global statistics file is being used for the regional.  
2. Based on usenewgfsberror, the arrays are dimensioned to handle the global dimensions in the regional run.
3. The additional variables in the global file are not used.  The sst values are read in and put  in the arrays, but are then overwritten with what is currently used in the regional runs.  So to use the global sst values, you should just be able to delete the additional code overwriting the stats.  For the ozone and cw, the test scripts I was given did not use these variables as control variables.  But I believe they can be done similarly.

After applying these changes, the global runs reproduced results when compared to the head of the trunk.  The regional did not.  A lot of time was spent trying to find the source of these differences.  A few errors in the original code were found.  These changes will probably change the regional fv3 results.  Even after these fixes, the results were different.  The differences appeared to be at the level of round-off errors.

jderber-NOAA:
The problem with the regional test was found to be that the same input file was being used for the output file.  Thus, the guess was changing with each run.  When this was fixed, the results for the regional input file was identical.

The changes are in branch fv3nam.

[...]

PR #199, opened 2021-08-13
 GitHub Issue NOAA-EMC/GSI#175. Use the global 127L B-Matrix in regional FV3 DA (FV3LAMDA)

jderber-NOAA:
…nal FV3 DA (FV3LAMDA)

This change allows the use of the global berror stats file in the regional runs.  Compared to the regional berror file, the global berror file is missing the second record that contains the latitudes and sigma's of the statistics.  Also, the dimensions of the arrays are different with the regional berror file containing some arrays that start with latitude 0 and end with latitude nlats+1.

To accommodate  the differences, a new global stats file was created which includes the second record containing the lats and sigmas.  This file is in /scratch1/NCEPDEV/da/John.Derber/git/gsi/berror_stats on Hera.  The global run will detect automatically whether the new or old format file is used and will run properly for either file.   For the regional, the default values will read the old regional berror file properly.  To use the new global file, the namelist parameter usenewgfsberror in bkgerr should be set to true.  Also, the code was fixed so that any file name could be used for the berror_stats file through berror_stats parameter in the namelist setup (with the default berror_stats).

### DIFF
--- a/src/gsi/balmod.f90
+++ b/src/gsi/balmod.f90
@@ -396,17 +396,6 @@ contains
 !   Set internal parameters to m_berror_stats
     call berror_set_reg('cwcoveqqcov',cwcoveqqcov)
 
-!   Read dimension of stats file
-    inerr=22
-    call berror_get_dims_reg(msig,mlat)
-
-!   Allocate arrays in stats file
-    allocate ( agvi(0:mlat+1,1:nsig,1:nsig) )
-    allocate ( bvi(0:mlat+1,1:nsig),wgvi(0:mlat+1,1:nsig) )
-    
-!   Read in background error stats and interpolate in vertical to that specified in namelist
-    call berror_read_bal_reg(msig,mlat,agvi,bvi,wgvi,mype,inerr)
-    
 !   ke_vp used to project SF to balanced VP
 !   below sigma level 0.8
 
@@ -419,12 +408,30 @@ contains
        endif
     enddo j_loop
     
-    agvk=zero
-    bvk=zero
-    wgvk=zero
     ke_vp=ke-1
     if (twodvar_regional) ke_vp=ke
-    if (.not.twodvar_regional) then
+
+!   Read dimension of stats file
+    inerr=22
+    call berror_get_dims_reg(msig,mlat)
+
+!   Allocate arrays in stats file
+    allocate ( agvi(0:mlat+1,1:nsig,1:nsig) )
+    allocate ( bvi(0:mlat+1,1:nsig),wgvi(0:mlat+1,1:nsig) )
+    
+!   Read in background error stats and interpolate in vertical to that specified in namelist
+    call berror_read_bal_reg(msig,mlat,agvi,bvi,wgvi,mype,inerr)
+
+!   Alternatively, zero out all balance correlation matrices
+!   for univariate surface analysis
+    if (twodvar_regional .or. lnobalance) then
+       if(mype==0) write(6,*)"***WARNING*** running univariate analysis." 
+       agvk=zero
+       bvk=zero
+       wgvk=zero
+       if(lnobalance) agvk_lm(:,:)=zero
+    else
+    
        do k=1,ke_vp
           do j=1,lon2
              do i=1,lat2
@@ -471,19 +478,8 @@ contains
           end do
        end do
     endif
-
-
-!   Alternatively, zero out all balance correlation matrices
-!   for univariate surface analysis
-    if (twodvar_regional .or. lnobalance) then
-       if(mype==0) write(6,*)"***WARNING*** running univariate analysis." 
-       bvk(:,:,:)=zero
-       agvk(:,:,:,:)=zero
-       wgvk(:,:,:)=zero
-       if(lnobalance) agvk_lm(:,:)=zero
-    endif
-    
     deallocate (agvi,bvi,wgvi)
+    
     
     return
   end subroutine prebal_reg
@@ -903,6 +899,7 @@ contains
 !$$$
     use kinds, only: r_single
     use gridmod, only: nlon,nlat,lat2,lon2,istart,jstart,region_lat
+    use m_berror_stats, only: berror_stats
     use constants, only: deg2rad,one
     implicit none
     
@@ -918,10 +915,9 @@ contains
 
 !   Read in dim of stats file
     lunin=22
-    open(lunin,file='berror_stats',form='unformatted')
+    open(lunin,file=berror_stats,form='unformatted')
     rewind lunin
     read(lunin)msig,mlat
-    
 
 !   Allocate and read in lat array in stats file
     allocate( clat_avn(mlat), clat_avn4(mlat) )
--- a/src/gsi/m_berror_stats.f90
+++ b/src/gsi/m_berror_stats.f90
@@ -43,9 +43,13 @@ module m_berror_stats
 
    private    ! except
 
-   ! reconfigurable parameters, via NAMELIST/setup/
-   public :: berror_stats    ! reconfigurable filename
+!   def usenewgfsberror - use modified gfs berror stats for global and regional.
+!                        for global skips extra record
+!                        for regional properly defines array sizes, etc.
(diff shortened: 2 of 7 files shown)


src/gsi/m_berror_stats.f90
```diff
@@ -125,6 +131,52 @@ subroutine get_dims(msig,mlat,mlon,lunit)
...
+   read(inerr,iostat=ier) msig,mlat,mlon_
+   call check_iostat(ier,myname_,'read header')
+   errtot=ier
+
+   errtot=0
+   allocate ( clat_avn(mlat) )
+   allocate ( sigma_avn(1:msig) )
+   read(inerr,iostat=ier)clat_avn,sigma_avn
+   do i=1,msig-1
+     if(sigma_avn(i) < sigma_avn(i+1))then
```
TingLei-NOAA, line 162: can some explanation be added here for why such a test could  determine the logical value of usenewgfsberror?

CatherineThomas-NOAA:
Will the new global background error with the additional entries be added to the fix directory as well?  We need to consider naming conventions if so to differentiate.  Another thing to consider is adding a utility to convert from one type of berror to the other.

src/gsi/m_berror_stats_reg.f90
```diff
@@ -807,103 +819,126 @@ subroutine berror_read_wgt_reg(msig,mlat,corz,corp,hwll,hwllp,vz,rlsig,varq,qopt
...
+        end do
+        do i=0,mlat+1
+           hwllp(i,loc)=hwll_default
+        end do
+     end if
   enddo
 
   deallocate(nrf3_loc,nrf2_loc,nmotl_loc)
+! Normalize vz with del sigmma and convert to vertical grid units!
+  if( .not. usenewgfsberror)then
```
TingLei-NOAA, line 926: what is the case for usenewgfsberror=.true. ?  I couldn't find this normalization of vz for it.

src/gsi/balmod.f90
```diff
@@ -419,12 +408,30 @@ subroutine prebal_reg(cwcoveqqcov)
        endif
     enddo j_loop
     
-    agvk=zero
```
CatherineThomas-NOAA, line 422: I think some of the zeroing out got improperly consolidated.  It looks like the first *vk=zero block is to initialize the variables before the loops.  The second block for zeroing out the *vk coefficients in the IF block is to turn the balance off.  I don't think that these can be combined.

src/gsi/m_berror_stats.f90
jderber-NOAA: Ting,

This test is just to see if the second record contains the expected values for the sigma_avn array.  It is possible, but very unlikely that if the wrong record is read, this criterion would be satisfied for all levels.

src/gsi/m_berror_stats_reg.f90
jderber-NOAA: You are correct, there is no normalization for vz when the new gfs berror file is used.  I could find no such normalization in the global runs and this result produced reasonable results.  If one is necessary, it can be added, but this whole code makes little sense to me.

src/gsi/balmod.f90
jderber-NOAA: I don't understand your concern.  There is nothing different from zeroing out the arrays for initializing the variable or for turning off the balance.  There is really no reason to initialize the variables to zero if they are then going to be overwritten.  I don't understand why you think they cannot be combined.

CatherineThomas-NOAA: Ah sorry, I missed the "else" and thought it was an "endif" for some reason and thought they were still being overwritten.  Ignore.

src/gsi/m_berror_stats_reg.f90
TingLei-NOAA: John, Thanks!  Did you mean this normalization is not necessary even for the "old" bkerr files ?  From the perspective of coding,  would it be preferable to have the same treatment to both the "old" and "new" bkerr files if there are no changes for the definitions of vz and related variables?

jderber-NOAA: I did not mean that it is not necessary for the old bkerr files.  In fact, as I think about it more, it probably should be there for the new one as well.  I could remove the check on usenewgffsberror = .true..  But with this change, the single obs plots should be produced to make sure that this produces reasonable results.  I cannot produce these plots.  Maybe we can have Xiaoyan produce these plots to make sure this change results in everything being OK.

TingLei-NOAA: John, Thanks.  Your clarification does make sense to me!

MichaelLueken-NOAA (commented):
@jderber-NOAA I have cloned your changes and ran the regression tests on WCOSS.  All tests successfully passed.  I did find a few issues while compiling and running the debug tests:

1) Two variables in m_berror_stats were flagged as unused (these have been noted in the review).  Please remove these unused variables.
2) The other issue was with the RTMA debug test.  On line 927 in m_berror_stats_reg, rlsig(1) and rlsig(2) are used.  However, the RTMA is a 2DVAR system, so rlsig(2) is out-of-bounds.  Please address this issue.

Once these corrections have been made, please use:

1) git add m_berror_stats.f90 m_berror_stats_reg.f90
2) git commit --amend
3) Please save without altering commit message
4) git push origin master --force

src/gsi/m_berror_stats_reg.f90
```diff
@@ -807,103 +819,126 @@ subroutine berror_read_wgt_reg(msig,mlat,corz,corp,hwll,hwllp,vz,rlsig,varq,qopt
...
+     dlsig(1)=rlsig(1)-rlsig(2)
```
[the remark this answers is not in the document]
jderber-NOAA: Mike,

I have the suggested changes, plus some small comment changes that were suggested by the reviewers.  I am afraid that I may have messed up the commit message because of some earlier changes that I made.

Johhn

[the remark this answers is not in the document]
jderber-NOAA: Please note that as Catherine noted we should include the version of the global berror stats containing the extra record.  Not sure what a good name would be.  Maybe include latsigma in the name.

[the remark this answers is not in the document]
MichaelLueken-NOAA: Hi John,

If you have committed changes that you have previously made without using:

`git commit --amend`

you can correct this using:

`git reset --soft HEAD~1`

This command will undo your previous commit, but keep the changes that were made.

You can use this command until you are back to the original commit, then use:

`git commit --amend`

If you changed the commit message using:

`git commit --amend`

You don't need to worry about it, so long as the GitHub issue number is present at the start of the commit message.

Please let me know if you have any questions.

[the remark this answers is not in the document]
jderber-NOAA: Mike,

I  guess I was doing things even more wrong.  I made my changes to the branch rather than my master.  I have now made the changes to the master and done the commit properly (I think).  I am rerunning test cases to ensure everything is OK with global (where I made some minor changes).  Will note here when completed and OK.

John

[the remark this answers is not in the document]
jderber-NOAA: Everything appears OK.

jderber-NOAA:
Mike,

Seems that the regional people want me to modify the vertical normalization for the global statistics for regional use.  I made these changes and put them into my master.  It only involved one change to the code and the removal of some comments.  Hope this is not too late.

MichaelLueken-NOAA:
@TingLei-NOAA and @CatherineThomas-NOAA Have you been able to look over the modifications that John made with respect to your comments?  Please double-check these changes at your earliest convenience and either give additional feedback or, if you are satisfied with the changes, please approve the changes.

TingLei-NOAA (approved):

CatherineThomas-NOAA (approved):

MichaelLueken-NOAA:
The due date for the review committee has past and additional reviewer feedback has been addressed.  I will now give final approval to this update and merge it to the authoritative master.

Issue #212, opened 2021-09-20
Updates to Correlated Error Code


KristenBathmann-NOAA:
Currently the correlated error code in the GSI src code cannot handle inconsistencies between the satinfo and the channels specified in a provided covariance file. This update will fix this, deleting row/columns in the covariance matrx if a channel is removed from the satinfo, or adding a row and column with no correlations if a channel is turned on in the satinfo. The changes will only occur in the initial setup routine of the correlated_obsmod, and no changes to runtime are expected.

KristenBathmann-NOAA:
Regression tests have passed, as expected. The only failure was global_fv3_4denvar_C192, but it was because of memory use. There were no differences between the control and update.

PR #217, opened 2021-09-28
GitHub Issue NOAA-EMC/GSI#212 updates to correlated error code

KristenBathmann-NOAA:
This update will allow for easier handling of correlated error in the case that an instrument has failures in active channels. If CrIS N20 looses the midwave channels, the only necessary change will be to the satinfo file. There is also the option now to turn on channels in the satinfo, without affecting correlated error.

If no changes are made to the satinfo, this update will not result in any difference to GSI results.

### DIFF
--- a/src/gsi/correlated_obsmod.F90
+++ b/src/gsi/correlated_obsmod.F90
@@ -444,7 +444,7 @@ logical :: corr_obs
       endif
       coun=0
       ErrorCov%R=zero
-      do ii=1,nctot !this should not be nctot, but something from satinfo KAB
+      do ii=1,nctot 
          if (iuse_rad(ii+istart)>0) then
             coun=coun+1
             ErrorCov%indxR(coun)=ii
--- a/src/gsi/correlated_obsmod.F90
+++ b/src/gsi/correlated_obsmod.F90
@@ -55,11 +55,11 @@ usually embedded in the {\it anavinfo} file. An example of such table follows:
 \begin{verbatim}
 correlated_observations::
 ! isis       method   kreq   kmut  type    cov_file
-  airs281_aqua  1      60.   1.0   ice     airs_rcov.bin
-  airs281_aqua  1      60.   1.0   land    airs_rcov.bin
-  airs281_aqua  1      60.   1.0   sea     airs_rcov.bin
-  airs281_aqua  1      60.   1.0   snow    airs_rcov.bin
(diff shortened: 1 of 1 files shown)


ADCollard (commented):
I'm a little confused about the comment on line 447: "!this should not be nctot, but something from satinfo KAB".  In fact, I agree it should not be nctot but an additional variable calculated from the loop starting at line 403?

HaixiaLiu-NOAA:
If I understand correctly, for an active channel that is in satinfo but not in covariance file, the covariances between this channel and other channels are zero and variance is set to use the number from the satinfo file. The function is good, however, it is not a good estimation of the covariances.  Should we always compute the covariances from OmB and OmA? I think it is okay to remove rows and columns that are in the covariance file, but not in the satinfo.

KristenBathmann-NOAA:
@ADCollard  Andrew, this was a debugging comment, and I will remove it. One thing that I am assuming in this update is that the total number of channels (monitor+active) as specified in the covariance file (as nctot) will be the same as it is in the satinfo. Setting iuse=-2 in the satinfo does not alter this count. Is it reasonable to assume that these counts will always be consistent? In the case that they differ (for example putting all 2211 CrIS channels into the satinfo) would require updates to the covariance file anyways.

KristenBathmann-NOAA:
@HaixiaLiu-NOAA Haixia, you are correct. Turning on a new channel in the satinfo will set cross-covariances to zero, and variance to the variance from the satinfo. It is best to compute covariances from OMB and OMA, but this offers a good starting point to generate those statistics. I know Erin Jones is planning some experiments with the CrIS shortwave channels, but does not want to turn off correlated error for the other channels initially.

HaixiaLiu-NOAA (approved):
thank you, Kristen. it is approved

HaixiaLiu-NOAA:
@KristenBathmann-NOAA Please merge the changes to the v16.x repo before you leave EMC.

CatherineThomas-NOAA:
@KristenBathmann-NOAA Have you run the regression tests for these changes?  If so, and if @ADCollard approves, I can do the merge.

ADCollard (approved):

KristenBathmann-NOAA:
@CatherineThomas-NOAA Yes I have run the regression tests.  The only failure was global_fv3_4denvar_C192, but it was because of memory use. There were no differences between the control and update.

CatherineThomas-NOAA (approved):

Issue #220, opened 2021-10-02
Update ConMon's MkBase script


EdwardSafford-NOAA:
The MkBase.sh script is used to generate the ConMon's historic values (baseline) file.  This file is used to generate warnings for substantially reduced observation counts in specific cycles.  

The MkBase.sh script is missing a feature which would enable it to evaluate only a single data source (i.e. t120) at a time, rather than evaluating all available sources (it's only mode of function at present).  This feature is in the RadMon's MkBase.sh script, which will be the template for this small fix.

PR #221, opened 2021-10-04
GSI issue #220.  Update MkBase utility.  Completes #220.

EdwardSafford-NOAA:
Updated MkBase utility adding the option of updating the historic obs counts of only a single data type (i.e. t120) instead of all data types.


MichaelLueken-NOAA:
There is no change to the source code, so I will now give final approval to these changes and merge them to the authoritative master.

PR #262, opened 2021-12-02
Change cp to mv for RadMon's warning files.

EdwardSafford-NOAA:
Changed cp to mv for RadMon's warning files (bad_*.txt).  The common workflow uses the same directory for RadMon and OznMon processing and the RadMon's bad_pen.${PDY} file was ending up in both the RadMon and OznMon $TANKdir locations.  Changing the RadMon's copy to a move fixes that.


MichaelLueken-NOAA:
Approving and merging updates to feature/gfsda.v16.1.5_wcoss2_port.

PR #376, opened 2022-05-05
GitHub Issue NOAA-EMC/GSI#373.  Compress RadMon data files.

### LINKED ISSUE #373: Compress RadMon tar files
Compress the RadMon's tar files which are used for storage of extracted data.

EdwardSafford-NOAA:
Make necessary changes to both data extraction and image generation in order to handle compressed RadMon data files.

### DIFF
--- a/util/Radiance_Monitor/data_extract/ush/radmon_copy.sh
+++ b/util/Radiance_Monitor/data_extract/ush/radmon_copy.sh
@@ -82,7 +82,7 @@ if [[ ${exit_value} -eq 0 ]]; then
 
    for type in ${type_list}; do 
 
-      file_list=`ls ${DATA_LOCATION}/${type}.*${PDATE}*ieee_d* `
+      file_list=`ls ${DATA_LOCATION}/${type}.*${PDATE}*ieee_d* ${DATA_LOCATION}/${type}*tar* `
 
       for file in ${file_list}; do
          bfile=`basename ${file}`
--- a/util/Radiance_Monitor/image_gen/ush/RadMon_IG_glb.sh
+++ b/util/Radiance_Monitor/image_gen/ush/RadMon_IG_glb.sh
@@ -154,7 +154,7 @@ fi
 
 if [[ ${pdate} -gt ${latest_data} ]]; then
   echo " Unable to plot, pdate is > latest_data, ${pdate}, ${latest_data}"
-  exit 4 
+  exit 5 
 else
   echo " OK to plot"
 fi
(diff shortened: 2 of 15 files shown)


EdwardSafford-NOAA:
@DavidHuber-NOAA please take a look if you have a moment.  Thanks!

util/Radiance_Monitor/data_extract/ush/radmon_copy.sh
```diff
@@ -82,7 +82,7 @@ if [[ ${exit_value} -eq 0 ]]; then
 
    for type in ${type_list}; do 
 
-      file_list=`ls ${DATA_LOCATION}/${type}.*${PDATE}*ieee_d* `
+      file_list=`ls ${DATA_LOCATION}/${type}.*${PDATE}*ieee_d* ${type}*tar* `
```
DavidHuber-NOAA: Should this be
`ls ${DATA_LOCATION}/${type}.*${PDATE}*ieee_d* ${DATA_LOCATION}/${type}*tar* ` or do the .tar* files possibly live in a different directory?

If not, then it's not necessary to include the tar* files in this list since you would be copying from and to the `$dest_dir`.

util/Radiance_Monitor/image_gen/ush/RadMon_IG_glb.sh
```diff
@@ -208,8 +208,8 @@ fi
 #
 nfile_src=`ls -l ${ieee_src}/*${PDATE}*ieee_d* | egrep -c '^-'`
 if [[ $nfile_src -le 0 ]]; then
-   if [[ -e ${ieee_src}/radmon_angle.tar ]]; then
-      nfile_src=`tar -tf ${ieee_src}/radmon_angle.tar | grep ieee_d | wc -l`
+   if [[ -e ${ieee_src}/radmon_angle.tar || -e ${ieee_src}/radmon_angle.tar.${Z} ]]; then
+      nfile_src=`tar -tf ${ieee_src}/radmon_angle.tar* | grep ieee_d | wc -l`
```
DavidHuber-NOAA, line 218: Is there any possibility that both `radmon_angle.tar` and `radmon_angle.tar.$Z` could both exist?  If so, this command will result in an error.  If there is a chance for that, then I'd suggest adding exception handling (if both: error) or add code to 'tell' the file contents of both.

util/Radiance_Monitor/image_gen/ush/RadMon_IG_glb.sh
```diff
@@ -251,7 +251,7 @@ fi
 #
 test_list=`ls ${ieee_src}/angle.*${PDATE}.ieee_d*`
 if [[ $test_list = "" ]]; then
-   test_list=`tar -tf ${ieee_src}/radmon_angle.tar | grep ieee_d` 
+   test_list=`tar -tf ${ieee_src}/radmon_angle.tar* | grep ieee_d`
```
DavidHuber-NOAA, line 260: Same as above.  More occurrences follow.

DavidHuber-NOAA (changes requested):

util/Radiance_Monitor/data_extract/ush/radmon_copy.sh
EdwardSafford-NOAA: Oh gosh yes, the $DATA_LOCATION is required there.

util/Radiance_Monitor/image_gen/ush/RadMon_IG_glb.sh
EdwardSafford-NOAA: I'm tempted to say it's not possible that both versions exist.  That's my intent.  But you're probably right to suggest some defensive measures just in case.  I'll flag the presence of both as an error condition.

util/Radiance_Monitor/image_gen/ush/RadMon_IG_glb.sh
EdwardSafford-NOAA: This won't be a problem -- if both a compressed and uncompressed version of the radmon_angle.tar file exist the check above will catch it and exit, so this won't be executed.

EdwardSafford-NOAA:
@DavidHuber-NOAA please take another look when you have a moment.  Thanks.

DavidHuber-NOAA (approved):
Looks good, @EdwardSafford-NOAA.  Thanks!

MichaelLueken:
These changes have been approved by @DavidHuber-NOAA and no changes have been no source code changes, so I will now give final approval to these changes and merge them to the authoritative repository.

PR #396, opened 2022-05-26
GitHub Issue NOAA-EMC/GSI#388. Verification of readin_localization=.true. for regional GSI.

### LINKED ISSUE #388: Verification of readin_localizaiion=.true. for regional GSI
One recent test of FV3-LAM GSI compiled with debug mode  using readin_localization=.true. shows error message about index exceeding the upper boundary in convert_km_to_grid_units (hybrid_ensemble_isotropic.F90) at the labelled line as shown below:
if(nz>1)then
do n=2,n_ens
nk=(n-1)*grd_loc%nsig
do k=1,grd_loc%nsig
s_ens_h_gu_x(nk+k)=s_ens_h_gu_x(k) !clt runtime error occured,
s_ens_h_gu_y(nk+k)=s_ens_h_gu_y(k)
enddo
enddo
endif
Further investigation let me suspect the assginments of s_ens_h_gu_x are not correctly done when a map should be done between (1:nsig) and (grd_loc%kbegin_loc:grd_loc%kend_alloc). 
readin_localization is not used in EMC's regional runs, but the vertically varied horizontal scale is regarded as beneficial as done in the global GSI. Hence, this issue is expected to invite comments/suggestions and document progress on this point

TingLei-daprediction:
A quick fix for an "exceeding array bound" error when regional GSI in debug mode ran with readin_localizaiion=.true..
See issue : https://github.com/NOAA-EMC/GSI/issues/388. 
The new code shows identical results when built with debug mode or optimization mode.
And, both of them are also identical with previous GSI built with optimization mode, which is expected when the involved array is stored continuously in the memory. 
The identical results are concluded according to their output in the gsi stdout, especially for their Final/Initial cost/gradient as shown below
image

### DIFF
--- a/src/gsi/hybrid_ensemble_isotropic.F90
+++ b/src/gsi/hybrid_ensemble_isotropic.F90
@@ -4175,7 +4175,7 @@ subroutine convert_km_to_grid_units(s_ens_h_gu_x,s_ens_h_gu_y,nz)
   implicit none
 
   integer(i_kind) ,intent(in   ) ::nz
-  real(r_kind),intent(  out) ::s_ens_h_gu_x(nz),s_ens_h_gu_y(nz)
+  real(r_kind),intent(  out) ::s_ens_h_gu_x(nz*n_ens),s_ens_h_gu_y(nz*n_ens)
   logical :: print_verbose
   real(r_kind) dxmax,dymax
   integer(i_kind) k,n,nk


ShunLiu-NOAA (approved):

hu5970 (approved):

MichaelLueken:
The due date for these changes has past with no feedback, so I will now give final approval to these changes and merge them to the authoritative repository.

PR #404, opened 2022-06-03
Prune non-global files.

aerorahul:
This PR adds an option to `enable/disable` pruning of extraneous files when preparing for NCO.  Default is `OFF`.

PR #405, opened 2022-06-03
bug in the if statement for pruning

aerorahul:
There is a bug in the pruning if statement from #404 that needs to be corrected.


MichaelLueken (approved):
Approving bug fix.

MichaelLueken:
Merging bug fix to `ush/build_4nco_global.sh` in `release/gfsda.v16.3.0`.

Issue #468, opened 2022-09-01
Enhancements to simultaneous multiscale EnVar using scale- and variable-dependent localization for the global and regional predictions and parallel ensemble IO for EnVar


Wangy1111:
The rapid advancement of new computational resources and technologies permits the global and regional NWP models to resolve a much wider range of scales. Such advancement also allows observations sampling a variety of scales to be ingested in the NWP models. Data assimilation (DA) algorithms, therefore, are required to appropriately analyze the state and its uncertainties across multiple scales.  Here we introduce a simultaneous multiscale DA approach in GSI EnVar. This multiscale DA approach is realized by utilizing model space scale-dependent localization to simultaneously update all resolved scales. Its mathematical demonstration and examination for the global application were presented in Huang et al. (2021). This multiscale approach was additionally developed to include variable-dependent localization and the extended approach was applied to assimilate radar observations for convective-scale forecasting in Wang and Wang (2022). The code developed aims to improve EnVar control analysis and subsequent forecasts for the global and regional models and includes the following aspects:

First, spatial scale-dependent localization is implemented in EnVar following Huang et al 2021.  The codes apply to both global and regional applications. 

Second, a variable-dependent localization method proposed by Wang and Wang (2022) is implemented in EnVar.  The codes apply to both global and regional applications. 

The codes are enhancements relative to #459.

In addition, parallel ensemble IO for both conventional and radar EnVar for FV3-LAM is implemented by reading in all ensemble members simultaneously.

PR #473, opened 2022-09-03
Simultaneous multiscale EnVar using scale- and variable-dependent localization for the global and regional predictions

Wangy1111:
This PR aims to add multiscale DA capability by using scale- and variable-dependent localization in EnVar and is related to issue #468 
These code developments apply to both regional and global applications. We have performed global and regional tests for checking the added capability.


Wangy1111:
We only access Orion and are unable to perform regression tests for this branch. We also would like to assign @dtkleist as a code reviewer. Thank you.

Issue #478, opened 2022-09-27
Incorrect analysis date in calc_analysis.x atmanl.nc file


RussTreadon-NOAA:
NCO testing of GFS v16.3.0 found that `pgrb2.1p00.anl` contained the wrong analysis time.    Investigation by EIB found the wrong analysis date encoded in the `atmanl.nc` file created by `calc_analysis.x`.   

The wrong `w3emc` library is used to build `calc_analysis.x`.   `CMakeLists.txt` incorrectly references 

   `target_link_libraries(calc_analysis.x PRIVATE w3emc::w3emc_d)`

This line should read

   `target_link_libraries(calc_analysis.x PRIVATE w3emc::w3emc_4)`

This issue is opened to document this error and its correction.

RussTreadon-NOAA:
Work on this issue will be done in forked branch [`RussTreadon-NOAA/GSI/release/gfsda.v16.3.0`](https://github.com/RussTreadon-NOAA/GSI/tree/release/gfsda.v16.3.0).

RussTreadon-NOAA:
**WCOSS2 GFS v16.3 test**

Set up v163ctl on Cactus.  Populate v163ctl ROTDIR with files for 2022092700 from NCO's real-time GFS v16.3 parallel.   Submit 2022092706 gfs prep followed by anal and analcalc.  For analcalc, add 

`export CALCANLEXEC=/lfs/h1/ops/para/packages/gfs.v16.3.1/exec/calc_analysis.x` 

to `config.analcalc`.   Confirm that the incorrect analysis date is written to `gfs.t06z.atmanl.nc`
```
russ.treadon@clogin05:/lfs/h2/emc/ptmp/russ.treadon/comrot/v163ctl/gfs.20220927/06/atmos> ncdump -hcs gfs.t06z.atmanl.nc |grep "time:units"
                time:units = "hours since 2045-03-02 00:00:00" ;
```
This is also seen, as @aerorahul pointed out, in `gfsanalcalc.log`
```
nid001419.cactus.wcoss2.ncep.noaa.gov 0:  Background initialization date=        2022           9          27           0
nid001419.cactus.wcoss2.ncep.noaa.gov 0:            0           0
 nlon=        3072
 nlat=        1536
 nlev=         127
 Analysis valid date=        2045           3           2           0
           0           0
nid001419.cactus.wcoss2.ncep.noaa.gov 0:  Copying from Background grid_xt
```

[...]

PR #479, opened 2022-09-27
Update calc_analysis.fd/CMakeLists.txt to use w3emc_4 (#478)

RussTreadon-NOAA:
Correct bug in the release/gfsda.v16.3.0 build of `calc_analysis.x`.    Please see #478 for details.

### DIFF
--- a/util/netcdf_io/calc_analysis.fd/CMakeLists.txt
+++ b/util/netcdf_io/calc_analysis.fd/CMakeLists.txt
@@ -17,7 +17,7 @@ target_link_libraries(calc_analysis.x PRIVATE MPI::MPI_Fortran)
 target_link_libraries(calc_analysis.x PRIVATE bacio::bacio_4)
 target_link_libraries(calc_analysis.x PRIVATE nemsio::nemsio)
 target_link_libraries(calc_analysis.x PRIVATE ncio::ncio)
-target_link_libraries(calc_analysis.x PRIVATE w3emc::w3emc_d)
+target_link_libraries(calc_analysis.x PRIVATE w3emc::w3emc_4)
 
 # Install executable targets
 install(TARGETS calc_analysis.x RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})


aerorahul (approved):
Thanks for the quick fix @RussTreadon-NOAA

aerorahul:
@RussTreadon-NOAA 
I can create a companion PR in `GSI-utils/develop`.

RussTreadon-NOAA:
@aerorahul , `GSI-utils/develop` is OK.   It uses `w3emc_4`.

https://github.com/NOAA-EMC/GSI-utils/blob/develop/src/netcdf_io/calc_analysis.fd/CMakeLists.txt

emilyhcliu:
@RussTreadon-NOAA Thanks for fixing the bug.  What was the impact of this bug?

emilyhcliu:
> @RussTreadon-NOAA Thanks for fixing the bug. What was the impact of this bug?

@RussTreadon-NOAA @aerorahul 
I read more e-mails related to this.  This PR fixed the erroneous time records found in the grib2 forecast files, and the source of error was from the calc_analysis.x.  Do I understand the issue and the fix correctly?

RussTreadon-NOAA:
> > @RussTreadon-NOAA Thanks for fixing the bug. What was the impact of this bug?
> 
> @RussTreadon-NOAA @aerorahul I read more e-mails related to this. This PR fixed the erroneous time records found in the grib2 forecast files, and the source of error was from the calc_analysis.x. Do I understand the issue and the fix correctly?

Yes, @emilyhcliu , your understanding of the problem is correct.  calc_analysis.x was being built with the wrong w3emc library.  As a result the analysis date calculation in calc_analysis.x generated the wrong value.   This was not due to a code problem but rather building with the wrong w3emc library.   This fix was simple:  build the calc_analysis.x code with the correct w3emc library.

emilyhcliu:
@RussTreadon-NOAA @aerorahul 
Does this bug fix only fix the time issue, and the rest of the analysis fields stay the same? 
In other words, does this bug fix change the analysis fields and, thus, the subsequent forecasts?
We also use w3emc_d in other parts of the GSI (per communication with Lin).  Is it sufficient to fix the calc_analysis.x only?

RussTreadon-NOAA:
The bug in question pertains specifically to calc_analysis.x.   Compiling calc_analysis with the correct w3emc library only alters the time written to the output analysis file, atmanl.nc.   The use of w3emc_d, w3emc_4, or w3emc_8 depends on which w3emc routines are used and which arguments are passed.   If you are concerned about which w3emc is used in various parts of the DA cycle, you can compare w3emc use in current operational DA apps with the use in gfsda.v16.3.0.   If you are concerned about the impact of the change in the w3emc used to build calc_analysis.x on the cycled system, we can run a parallel with the bug and compare output with Lin's corrected parallel.

emilyhcliu:
@RussTreadon-NOAA Thanks for your explanation.  I understand that switching to w3emc_4 for calc_analysis only alters the time stamp.  Based on my understanding, I do not think we need to run with the bug to compare with Lin's corrected parallel.

PR #481, opened 2022-10-03
Accept a local configuration file for regression tests

venitahagerty:
When running regression tests, users on Hera can set an environment variable config_path to the location of a shell local_vars.sh that contains statements to set environment variables. If this option is not used, current values should be set as defaults. When it is used, a user can override some path names and also variables such as group, queue, and accnt. This allows a wider range of users to run the regression tests without editing shells.


venitahagerty:
Requesting review from @aerorahul , @dtkleist , and @RussTreadon-NOAA

RussTreadon-NOAA:
@venitahagerty , I cloned your branch on Hera, built gsi.x, and tried the following.

**Check 1 - do not define `config_path`**   
If the user does not define `config_path`, ctest fails when executing `regression_var.sh`.   For example, `ctest -VV -R global_T62_ozonly` yields
```
test 2
    Start 2: [=[global_T62_ozonly]=]

2: Test command: /scratch1/NCEPDEV/da/Russ.Treadon/git/gsi/pr481/regression/regression_driver.sh "global_T62_ozonly" "/scratch1/NCEPD
EV/da/Russ.Treadon/git/gsi/pr481/build/regression"
2: Test timeout computed to be: 86400
2: + export regtest=global_T62_ozonly
2: + regtest=global_T62_ozonly
2: + '[' -d ']'
2: + source /local_vars.sh
2: /scratch1/NCEPDEV/da/Russ.Treadon/git/gsi/pr481/regression/regression_driver.sh: line 10: /local_vars.sh: No such file or director
y
1/1 Test #2: [=[global_T62_ozonly]=] ..........***Failed    0.00 sec
```

Does this PR assume developers will always set `config_path`?    

If not, we could test if `config_path` is set prior to the `config_path` directory test.  I added a `-z` check to `regression_driver.sh` 
```
# option to set local values with a config file
if [ ! -z "$config_path" ]; then
    if [ -d $config_path ]; then
        source $config_path/local_vars.sh
    fi
fi
```
This allowed ctest to run without defining `config_path`.

**Check 2 - use user configuration**
I created `local_vars.sh` and populated it with
```
queue="debug"
ptmp="/scratch1/NCEPDEV/stmp2/$LOGNAME/gsi_regtest"
accnt="fv3-cpu"
```
I set `config_path` to the directory in which `local_vars.sh` resided and executed the same ctest command above.  This worked.  The ctest finished more quickly (304.38 seconds) given the shorter debug queue wait times compared to running the same ctest in the batch queue (723.89 seconds).

Depending on how the refactoring of gsi/enkf ctests evolves, more variables in `regression_var.sh` could be added to the list of variables configurable via `$config_path/local_vars.sh`.

christinaholtNOAA:
@RussTreadon-NOAA I'm curious Check 1 might be the result of a bash oversight. We may get the desired check without the additional logic if we just add quotes. I ran this toy case with the altered logic:

```
$ foo=
$ if [ -d "$foo" ] ; then echo foo ; fi
$
$ foo=../
$ if [ -d "$foo" ] ; then echo foo ; fi
foo
$ unset foo
$ if [ -d "$foo" ] ; then echo foo ; fi
$
```

RussTreadon-NOAA:
@christinaholtNOAA , you are right.   I added double quotes around `"$config_path"`.   The ozone ctest ran OK without `config_path` being defined in my environment.   Of course, it still ran when `config_path` was defined.   Your solution is better than what I suggested.

venitahagerty:
Added double quotes around "$config_path" to fix that bug.

RussTreadon-NOAA (commented):
The proposed changes have been tested with and without setting local environment variable `config_path` and with and without `local_vars.sh` present.    

- Default behavior is observed for ctest when `config_path` is not defined.  
- When `config_path` is defined and `$config_path/local_vars.sh` exists, the ctests run with user changes picked up from `local_vars.sh`.   
- When `config_path` is defined but `local_vars.sh` does not exist, ctest fails.   Adding `-V` to the ctest command confirms that the failure is due to `$config_path/local_vars.sh` not existing.

This PR adds optional flexibility without changing default behavior. 

This PR is currently 3 commits behind develop.   Please update the forked `feature/configure_ctests` to the head of develop.

venitahagerty:
Fork is now updated to the head of develop.

RussTreadon-NOAA:
Thank you, @venitahagerty , for updating your branch to the head of GSI develop.  I updated my working copy of your branch and ran the `global_T62_ozonly` ctest three times:
- do not define `config_path` (default behavior).   This ran as expected but it took 2643.81 seconds to run four jobs since the default configuration uses the batch queue.
- define `config_path` but do not provide `local_vars.sh`.   Rerunning with `ctest -V` clearly identified the missing `$config_path/local_vars.sh` as _No such file_.  
- define `config_path` and provide `local_vars.sh`.   This ran as expected with my settings used.   I specified the debug queue.  The four jobs in `global_T62_ozonly` completed in 303.67 seconds.

RussTreadon-NOAA (approved):
Approve.   It would be good to get reviews from other GSI developers and/or the EMC Engineering & Implementation Branch.

venitahagerty:
How can I get additional reviews and approvals?

RussTreadon-NOAA:
> How can I get additional reviews and approvals?

Usually the developer identifies two peer reviewers, reaches out to them to see if they can review the PR, and assigns to those who agree to review the PR.   One additional review is required from a staff member who has permission to merge branches into the authoritative repository develop.   I can act as the additional reviewer.   Are there GSI developers you can contact at GSL to review your PR?

CoryMartin-NOAA (approved):
Changes seem straightforward enough to me.

RussTreadon-NOAA:
> Changes seem straightforward enough to me.

Thank you, @CoryMartin-NOAA !

RussTreadon-NOAA:
@venitahagerty , I will merge your PR into develop this morning.   

For future GSI PRs please begin by opening a [GSI issue](https://github.com/NOAA-EMC/GSI/issues).  The procedure for making changes to NOAA-EMC/GSI develop are not well documented in the [github GSI wiki](https://github.com/NOAA-EMC/GSI/wiki).   This information currently resides in VLAB under [GSI code management](https://vlab.noaa.gov/redmine/projects/comgsi/wiki/Change_Control_management_for_the_GSI).   Moving this information into the github GSI wiki is on our to-do list.

PR #484, opened 2022-10-05
Allocate fhour in write_sfc_ prior to use (#483)

### LINKED ISSUE #483: BUG FIX:  allocate fhour in write_sfc_ prior to use
`gsi.x` seg faults when writing netcdf file `sfcanl.gsi`.   A check of `netcdfgfs_io.f90` found that subroutine `write_sfc_` does not allocate local array `fhour` before use.    `gsi.x` runs to completion after adding a line to allocate `fhour` prior to use

This issue is opened to document this bug and its fix.

RussTreadon-NOAA:
This PR is opened to request merger of a bug fix for `write_sfc_` (file `src/gsi/netcdfgfs_io.f90`).   The version of `write_sfc_` in the authoritative repository does not allocate local array `fhour` prior to use.   This bug causes `gsi.x` to seg fault when writing the netcdf `sfcanl_gsi` output file.   The fix in this PR is simple:  allocate `fhour` prior to use.    

Merger of this PR into the authoritative develop will close issue #483.


CatherineThomas-NOAA:
@RussTreadon-NOAA  I've taken a look at the code.  The fix is straightforward.  To double check my understanding: this bug is technically in ops/our parallels but the routine in question is not exercised when NSST is on.  Is this correct?

**Edit: Ignore that, I followed the wrong IF blocks.

emilyhcliu (approved):

RussTreadon-NOAA:
> @RussTreadon-NOAA I've taken a look at the code. The fix is straightforward. To double check my understanding: this bug is technically in ops/our parallels but the routine in question is not exercised when NSST is on. Is this correct?
> 
> **Edit: Ignore that, I followed the wrong IF blocks.

@CatherineThomas-NOAA , you are right.   GFS v16.2 and the to-be-implemented GFS v16.3 do not reach the modified code in this PR.   When the operational GFS atmos_anlaysis job runs `gsi.x` it is configured such that it writes `dtfanl.nc`.

CatherineThomas-NOAA:
Thanks for confirming @RussTreadon-NOAA, I managed to majorly confuse myself there.

It looks like that this would also occur in the subroutine `write_sfc_nst_` in the same file.  Another allocation should probably be added there as well (~line 2754).

RussTreadon-NOAA:
@CatherineThomas-NOAA , you are right.   Good catch!  The same issue occurs on line 2754 of `netcdfgfs_io.f90`.    Allocatable array `fhour` is initialized to zero before being allocated.   Both write_sfc routines, `write_sfc_` and `write_sfc_nst_`, in `netcdfgfs_io.f90` have the same problem.    `bugfix/sfcanl` has been updated to allocate `fhour` prior to use in `write_sfc_nst_`.  Done at [b1df681](https://github.com/RussTreadon-NOAA/GSI/commit/b1df681eefeb578526a7de38adaa1267fe820062).

CatherineThomas-NOAA (approved):
Looks good to me now.  Approved.

CoryMartin-NOAA:
After @CatherineThomas-NOAA 's approval, I think this looks ready to merge given all checks have passed.

PR #489, opened 2022-10-13
Fix fortran standard non-conforming code for rerank interface

PeixinQiao:
According to Fortran 2018 standard Section 15.4.3.2 point 7, if an explicit specific interface for an external procedure is specified by and interface body or a procedure declaration statement, the characteristics shall be consistent with those specified in the procedure definition.

`m_rerank.f90` is standard non-conforming and one small reproducer is in https://github.com/llvm/llvm-project/issues/57620#issuecomment-1275525166. The test results show both `gfortran` and `ifort` do not the expected behaviors, which is to rerank the array pointer. One possible fix is in https://github.com/llvm/llvm-project/issues/57620#issuecomment-1277138294 and `gfortran`, `ifort` and LLVM Flang support it.

### DIFF
--- a/src/gsi/m_rerank.f90
+++ b/src/gsi/m_rerank.f90
@@ -3,6 +3,7 @@ module m_rerank
 ! 01Aug2011 - Lueken  - changed F90 to f90 (no machine logic
 ! 28Apr2011 - Todling - overload to handle single precision
   use kinds, only : r_double,r_single,i_kind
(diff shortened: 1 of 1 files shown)


PeixinQiao:
@RussTreadon-NOAA @MichaelLueken

RussTreadon-NOAA (approved):
Changes reviewed.  Look good.  Approve.

Issue #500, opened 2022-11-22
The issue with RRFS (FV3-LAM) GSI EnKF runs on the north american domain.  


TingLei-daprediction:
The RRFS (FV3-LAM based ) hybrid GSI DA  is targeting to run on the 3km north american domain (model grids: 3950X2700X65) . Current tests show that make necessary  further ,sometimes significant, optimization of codes to handle the resulted large memory requirements.  This issue would document the issues/problems found so far for current EnKF runs. 
The current EnKF would abort with 5 control variables (even one control) variables with 30 or 2 ensembles.  Investigation with EnFK built with debug mode give the following information:
the program would fail for sendbuf in loadbal.f90 . 
The issue was found to potentially come from two: 1 . the size of senbuf is beyond the recognition of the (i_kind) integer (also for int64 integer) 2. such "large " index access to the sendbuf arrays. 
The attached is a discussion among us with more details 
^^

     In running Shun's NA EnKF case. with debug mode, it points to the line 403         in loadbal.f90
          ^^
              sendbuf(n) = grdin(indxproc(np,i),nn,nb,ne)
          VV
      Further tracing points to the definition of sendbuf 
            ^^
                allocate(sendbuf(numproc*nanals_per_iotask*npts_max*ncdim))
            VV
            The numproc, , nanals_per_iotask, npts_max, ncdim are , in this case,                160,1,66792,326,
            in which 160*66792 approximately equal to the total number of grid
            points in the horizontal dir for one member. 
            Hence, it could n't be changed by more cpu/nodes or smaller ensemble sizes.
           Their multiplication gets a big number beyond what the i_kind type
           format  integer could recognize. 
           However, It seems the allocation still works .
           But the problem would occur in the above line
           when the n became so large that can't be recognized too  and, "large:                  index access to array would also cause segmentation error.  
          I have a small test code for you to play with 
             ^^
                    program test
                        integer nn
                       real,allocatable :: ar (:)

                      nn=160*66792*326
                      allocate(ar(160*66792*326)). !ok
                      ar(160*66792*326)=1.0 ! segmentation error
                       write(6,*)'nn is' ,n.  !  a wrong number (negavie value) is given
                       end 
                 VV
VVVV
This issue would document the issues and solutions in progress , which has been worked on by a few EMC developers, collaborators and WCOSS2 help experts.  I would cc here some of them: @jswhit2  @ShunLiu-NOAA @hu5970  @JacobCarley-NOAA @dtkleist

[...]

PR #501, opened 2022-11-22
EnKF updates to support RRFS 3-km NA domain

jswhit:
WIP:  see issue #500


RussTreadon-NOAA:
@jswhit2 , thank you for quickly opening this PR in response to issue #500.   

 @TingLei-NOAA , since you opened #500, would you please review this PR?   Do the changes in this PR resolve the challenges described in #500.  Who else do you recommend as a reviewer?

TingLei-NOAA:
@RussTreadon-NOAA  Sure. My pleasure to be as one reviewer for this PR.  I would recommend @hu5970 as another reviewer. 
Thanks a lot to @jswhit . This PR has resolved most of abortions we saw on wcoss2 and, I believe, it has, at least basically, done the EnKF optimization needed for runs on NA .    I will report the results  with this PR  on WCOSS2 and Hera .

RussTreadon-NOAA:
> @RussTreadon-NOAA Sure. My pleasure to be as one reviewer for this PR. I would recommend @hu5970 as another reviewer. Thanks a lot to @jswhit . This PR has resolved most of abortions we saw on wcoss2 and, I believe, it has, at least basically, done the EnKF optimization needed for runs on NA . I will report the results with this PR on WCOSS2 and Hera .

Thank you, @TingLei-NOAA .   Added @hu5970 as second reviewer.

TingLei-NOAA (approved):

TingLei-NOAA:
The current PR passed my own regression test for fv3-lam (generated the identical results from their standard output), though the GSI regression tests are not finished yet .  All changes look good to me! 
The GSI EnKF resolved most segmentation issues we saw on wcoss2 and its' successful run of EnKF (2 member ) on the north american domain had been repeated.  All means this PR for GSI EnKF attains a big improvement in dealing with large domains and is approved.    However, when we are still working on the EnKF using this version on the north american domain on wcoss2 , I would suggest we won't close the associated issue(#500)  when this PR is done.

RussTreadon-NOAA:
@TingLei-NOAA , did the standard suite of regression tests pass for the changes in this PR?   If so, please update this PR with this information and re-approve.  Thanks!

hu5970 (dismissed):
The code looks good.

TingLei-NOAA:
@RussTreadon-NOAA I am having some problems running the existing GSI regression tests and I am debugging it (not about GSI codes but about the configure and scripts for the regression tests).    I will update here if any findings.  And, that will be great if you could show me an example (including procedures) of GSI regression tests for current GSI.

RussTreadon-NOAA:
@TingLei-NOAA , at present GSI regression tests only run on Hera.   [GSI Ctests (regression tests)](https://github.com/NOAA-EMC/GSI/wiki/GSI-Ctests-(regression-tests)) describes how to set up and run the regression tests on Hera.   Ctest documentation is a work in progress.  Your walking through the steps will help us improve the documentation.

TingLei-NOAA:
@RussTreadon-NOAA Got it!  Would you please open an issue for GSi regression tests and hence, all of us could add our comments/questions to it?

RussTreadon-NOAA:
@TingLei-NOAA , very good idea.  Please open the issue you mention.

RussTreadon-NOAA:
@TingLei-NOAA , I ran the suite of 19 regression tests using the current head of `develop` as the control the `jswhit:fv3regfixes` as the update.    18 of the 19 regression tests passed.    

global_fv3_4denvar_C192 failed.  Examination of the summary file for global_fv3_4denvar_T126 shows that the failure was due to a scalability test.  This is not a fatal fail.   A rerun of global_fv3_4denvar_T126 passed.

TingLei-NOAA:
@RussTreadon-NOAA Thanks a lot.  I had finished a round of ctests too!  some failed. but their ".out" files were deleted by the default cleaning option and hence, I am re-running ctest :(.   So, that is great you confirm the PR pass existing regression tests!
Please let me know if I could be of any further assistance.

RussTreadon-NOAA:
@TingLei-NOAA , all we need you to do is complete your review of Jeff's changes.   This can be done once your rerun of the regression tests finishes with acceptable results.

TingLei-NOAA:
@RussTreadon-NOAA Since you have finished the existing GSI regressions tests for this PR.  I would re-approve it. 
Thanks.

TingLei-NOAA (dismissed):

TingLei-NOAA:
This PR had been verified by FV3REG EnKF as reported before and @RussTreadon-NOAA also confirm it passes all existing GSI regression tests.  Approved.

RussTreadon-NOAA:
Thank you, @TingLei-NOAA  for the approval.   I will not run the regression tests on future PRs for which I am not a reviewer.

RussTreadon-NOAA:
Thank you, @TingLei-NOAA  for the approval.   I will not run the regression tests on future PRs for which I am not a reviewer.

src/enkf/controlvec.f90
```diff
@@ -309,11 +309,21 @@ subroutine write_control(no_inflate_flag)
          print *,'--------------'
       endif
       ! gather ensmean increment on root.
-      do ne=1,nanals_per_iotask
-         call mpi_reduce(grdin(:,:,nb,ne), grdin_mean_tmp, npts*ncdim, mpi_real4,&
-                         mpi_sum,0,mpi_comm_io,ierr)
-         if (nproc == 0) grdin_mean(:,:,nb,1) = grdin_mean(:,:,nb,1) + grdin_mean_tmp
-      enddo
+      if (real(npts)*real(ncdim) < 2**32/2. - 1) then
```
RussTreadon-NOAA: Need to specify numeric precision for `2**32/2. - 1` (GSI code  standard)

src/enkf/loadbal.f90
```diff
@@ -356,75 +355,128 @@ subroutine scatter_chunks
...
-        enddo
-     end do
-  end do
-  !$omp end parallel do
-
-enddo ! loop over nbackgrounds
-
 !==> compute mean, remove it from anal_chunk
-!$omp parallel do schedule(dynamic,1)  private(nn,i,n,nb)
+!!$omp parallel do schedule(dynamic,1)  private(nn,i,n,nb)
```
RussTreadon-NOAA: Do we plan on threading this loop at a later date?  If not, we should remove lines 479 and 493 since the extra `!` results in the omp directive being commented out.

jswhit2: That openmp loop was turned off to avoid a stack overflow segfault.

src/enkf/controlvec.f90
jswhit2: I will fix this.

src/enkf/loadbal.f90
```diff
@@ -460,52 +511,94 @@ subroutine gather_chunks
...
-  call mpi_alltoallv(sendbuf, scounts, displs, mpi_real4, recvbuf, rcounts, displs,&
-                     mpi_real4, mpi_comm_world, ierr)
-  if (nproc <= ntasks_io-1) then
-    do np=1,numproc
-     do ne=1,nanals_per_iotask
-      do nn=1,ncdim
-       do i=1,numptsperproc(np)
-         n = ((np-1)*ncdim*nanals_per_iotask + (ne-1)*ncdim + (nn-1))*npts_max + i
-         grdin(indxproc(np,i),nn,nb,ne) = recvbuf(n)
+if (real(numproc)*real(nanals_per_iotask)*real(npts_max)*real(ncdim) < 2**32/2. - 1) then
```
RussTreadon-NOAA: Same comment as for `controlvec.f90`.   One option is to replace `1` and `2` with `one` and `two` from the constants module.   That is, `2**32/2. - 1` becomes `two**32/two - one`.

RussTreadon-NOAA (commented):
Changes look good.   Code compiled and run on WCOSS2 using C48 regression tests.   No change in analysis results with respect to develop.   Two minor comments.  One for conformance to GSI coding standards.  The other asking if commented out omp directives should be removed.

src/enkf/loadbal.f90
RussTreadon-NOAA: Got it.   We should remove the omp directives to prevent someone from seeing these lines, reactivating omp for this loop, and encountering a segfault.

src/enkf/loadbal.f90
RussTreadon-NOAA: Ignore my bad suggestion.  `two**32/two -one` does not work.  Using this expression alters the analyses. 
  `2_i_kind**32/2._real - 1_i_kind` works.  This expression yields analysis files that are bitwise identical with the original code.

TingLei-NOAA: @RussTreadon-NOAA Thanks for investigation on it.  It is interesting, why "two**32/two-one"  alters the analyses. It reminds me that my verification for fv3-reg doesn't test the new scheme ( only invoked the "old" procedure decided by evaluation here using  (2**31-1) )  . I will rerun a verification by forcing use of the new procedure brought up by this PR .  I will finish it not later than tomorrow .

jswhit2: I used r_kind instead of real, hopefully that works

src/enkf/loadbal.f90
jswhit2: re-ran the test with omp back on and it worked, so uncommented those lines

src/enkf/controlvec.f90
RussTreadon-NOAA: @TingLei-NOAA 's comment prompted me to add print statements to see which branch of the `if-then` block is entered.  Much to my surprise, the `else` branch was always entered despite the fact that the product of `npts` and `ncdim` is `7,526,424`.     This shouldn't be the case since `2*32/2.-1` is `2,147,483,647`

I wrote a simple test code
```
program test
  integer(4),parameter:: one_i4 = 1_4
  integer(4),parameter:: two_i4 = 2_4
  real(8),parameter:: one_r8= 1.0_8
  real(8),parameter:: two_r8 = 2.0_8
  real(8):: term

  term = 2**32
  write(6,*)'2**32 is is ',term

  term = 2**32/2.
  write(6,*)'2**32/2. is ',term

  term=2**32/2. - 1
  write(6,*)'2**32/2. - 1 is ',term

  term=two_i4**32/two_r8 - one_i4
  write(6,*)'two_i4**32/two_r8 - one_i4 is ',term

  term=two_r8**32/two_r8 - one_r8
  write(6,*)'two_r8**32/two_r8 - one_r8 is ',term

  term=two_r8**32/two_i4 - one_i4
  write(6,*)'two_r8**32/two_i4 - one_i4 is ',term          

end program test
```
Execution of the resulting `./a.out` yields
```
 2**32 is is   0.000000000000000E+000
 2**32/2. is   0.000000000000000E+000
 2**32/2. - 1 is   -1.00000000000000     
 two_i4**32/two_r8 - one_i4 is   -1.00000000000000     
 two_r8**32/two_r8 - one_r8 is   2147483647.00000     
 two_r8**32/two_i4 - one_i4 is    2147483647.00000  
```
This makes sense.  `2**32` is `4,294,967,296`.   But `2**32` in the code is integer arithmetic.  The maximum value for an integer(4) is `2,147,483,647`.

If we want the logical tests in `loadbal.f90` and `controlvec.f90` to work, we need to use real arithmetic to compute `2*32/2. - 1`.    It seems that as is the code is not working as intended.

What do you think, @jswhit2 ?

jswhit2: I've now updated the if/else checks to use ```2_r_kind**32/2_r_kind - 1_r_kind```

RussTreadon-NOAA: Recompile the code with the changes committed at dfd4e7dcc.   Unfortunately, a rerun `enkf.x` for the test case yields the following printout in `stdout`
```
ens. mean anal. increment min/max  v   -271.803466797        325.930572510
ens. mean anal. increment min/max  tv   -466.174316406        428.788879395
ens. mean anal. increment min/max  q   -198.692459106        204.148773193
ens. mean anal. increment min/max  oz   -915.857604980        1035.82275391
ens. mean anal. increment min/max  ps   -20.3442897797        19.5052604675
```
The control built from the head of develop has
```
ens. mean anal. increment min/max  u   -8.92652511597        11.2560396194
ens. mean anal. increment min/max  v   -8.99013900757        7.72295475006
ens. mean anal. increment min/max  tv   -4.58017873764        4.80190753937
ens. mean anal. increment min/max  q  -0.546840786934       0.393992841244
ens. mean anal. increment min/max  oz  -0.987113935480E-06   0.857482405081E-06
ens. mean anal. increment min/max  ps   -2.43661499023        2.57639765739
```
A query of the netcdf analysis increment files from the test code confirms that the min/max increments are unrealistically large.

Something isn't correct when execution enters the `if` branch of the 2*32 `if-then` block.  I am running on WCOSS2.   Does the code execute as expected on Hera or Orion?

jswhit2: Thank you for identifying this problem @RussTreadon-NOAA.  I will investigate further as soon as I get a free moment.  When the regression tests passed, the code was always going into the other branch of the if/else (the one intended for very large grids), is that correct?

TingLei-NOAA:
Yes. I confirm  @Russ Treadon - NOAA Federal ***@***.***> 's
results.  On hera,  when I forced the use of the  "old" one,
I also see obviously large increments.    So , seems the new "option
branch"   has been working fine, passing various regression tests.



______________________________

Ting Lei

Physical Scientist, Contractor with Lynker in support of

EMC/NCEP/NWS/NOAA

5830 University Research Ct., Cubicle 2765

College Park, MD 20740

***@***.***

301-683-3624

src/enkf/controlvec.f90
RussTreadon-NOAA: Yes @jswhit2 , I added print statements in each branch of the if-then block.   

- When we had `2**32/2. - 1`, execution always entered the else branch.    Analysis increments look reasonable.   
- After moving to `r_kind` arithmetic execution now enters the if branch.    Analysis increments are unphysically large.

src/enkf/loadbal.f90
```diff
@@ -356,73 +355,126 @@ subroutine scatter_chunks
...
 ! displs is displacement into send array for data to go to proc np
-do np=0,numproc-1
-   displs(np) = np*nanals_per_iotask*npts_max*ncdim
-enddo
-if (nproc <= ntasks_io-1) then
-   scounts = nanals_per_iotask*npts_max*ncdim
+
+if (real(numproc)*real(nanals_per_iotask)*real(npts_max)*real(ncdim) < 2_r_kind**32/2_r_kind - 1_r_kind) then
+    do np=0,numproc-1
+       displs(np) = np*nanals_per_iotask*npts_max
```
TingLei-NOAA: @jswhit  @RussTreadon-NOAA , seems the line 369 missed ncdim and should be 
scounts = nanals_per_iotask*npts_max*ncdim   ( as line 367 in the previous version). 
After this change, the enkf with this "option" branch  gave the identical results (viewed from the std output)

RussTreadon-NOAA: Good find @TingLei-NOAA .   I added `ncdim` to line 369, recompiled, and reran on WCOSS2.   This run reproduced results from `enkf.x` built from develop.

jswhit2: fixed now

TingLei-NOAA (approved):
I rebuilt the PR with @jswhit recent changes and the global_enkf_T62 regression test passed. So I re-approve this PR in case it is needed.

RussTreadon-NOAA (approved):
WCOSS2 tests run at C48 and C384.   `enkf.x` built from fv3regfixes reproduces output from `enkf.x` built from develop.

RussTreadon-NOAA:
Given two approvals I will merge this PR into develop.    As requested by @TingLei-NOAA issue #500 will remain open.

PR #536, opened 2023-02-16
Add dDZ for EnKF

### LINKED ISSUE #534: Add functions for conducting EnKF dBZ analysis in RRFS
Several functions added for dBZ analysis in EnKF:
1, Add dBZ ncdiag output for EnKF analysis
2, Enhanced binary dBZ diag output for EnKF analysis
3, Fix spurious analysis increments when assimilating reflectivity (from OU MAP)
4, Added EnKF interface for read in dBZ from FV3LAM ensemble forecast (from OU MAP)

Fixed a bug for read in subdomain surface  restart files

chunhuazhou:
**Description**

This update includes several functions added for dBZ analysis in EnKF:
1, Add dBZ ncdiag output for EnKF analysis
2, Enhanced binary dBZ diag output for EnKF analysis
3, Fix spurious analysis increments when assimilating reflectivity (from OU MAP)
4, Added EnKF interface for read in dBZ from FV3LAM ensemble forecast (from OU MAP)

Fixed a bug for read in subdomain surface restart files

Fixes #534

### DIFF
--- a/src/gsi/gsi_rfv3io_mod.f90
+++ b/src/gsi/gsi_rfv3io_mod.f90
@@ -2031,7 +2031,8 @@ subroutine gsi_fv3ncdf2d_read(fv3filenamegin,it,ges_z,ges_t2m,ges_q2m)
           end do
        end do
 
-       if(allocated(sfc1) .and. allocated(sfc))deallocate (dim_id,sfc,sfc1,dim)
+       if(allocated(sfc1) .and. allocated(sfc)) deallocate (sfc,sfc1)
+       if(allocated(dim)) deallocate (dim)
        if(allocated(sfc_fulldomain)) deallocate (sfc_fulldomain)
     endif  ! mype
 
--- a/src/gsi/gsi_rfv3io_mod.f90
+++ b/src/gsi/gsi_rfv3io_mod.f90
@@ -538,7 +538,7 @@ subroutine read_fv3_files(mype)
 ! Declare local variables
     logical(4) fexist
     character(6) filename
-    character(14) filenames
+    character(19) filenames
     integer(i_kind) in_unit
     integer(i_kind) i,j,iwan,npem1
     integer(i_kind) nhr_half
@@ -573,11 +573,19 @@ subroutine read_fv3_files(mype)
        in_unit=15
        iwan=0
 !WWWWWW setup for one first guess file for now
-      do i=0,9 !place holder for FGAT
+       do i=0,9 !place holder for FGAT
           if ( i == 6 ) then
-            write(filenames,"(A11)") 'fv3_dynvars'
+             if(fv3_io_layout_y > 1) then
+                write(filenames,"(A16)") 'fv3_dynvars.0000'
+             else
+                write(filenames,"(A11)") 'fv3_dynvars'
+             endif
           else
-            write(filenames,"(A12,I2.2)") 'fv3_dynvars_',i
+             if(fv3_io_layout_y > 1) then
+                write(filenames,"(A17,I2.2)") 'fv3_dynvars.0000_',i
+             else
+                write(filenames,"(A12,I2.2)") 'fv3_dynvars_',i
+             endif
           endif
           INQUIRE(FILE=filenames, EXIST=fexist)
           if(.not.fexist) cycle
@@ -1119,7 +1127,7 @@ subroutine read_fv3_netcdf_guess(fv3filenamegin)
         if (allocated(fv3lam_io_dynmetvars2d_nouv)) &
           write(6,*)' fv3lam_io_dynmetvars2d_nouv is ',(trim(fv3lam_io_dynmetvars2d_nouv(i)), i=1,ndynvario2d)
         if (allocated(fv3lam_io_tracermetvars2d_nouv))&
-          write(6,*)'fv3lam_io_dynmetvars2d_nouv is ',(trim(fv3lam_io_dynmetvars2d_nouv(i)),i=1,ntracerio3d)
+          write(6,*)'fv3lam_io_tracermetvars2d_nouv is ',(trim(fv3lam_io_tracermetvars2d_nouv(i)),i=1,ntracerio2d)
       endif      
 
       if (laeroana_fv3cmaq) then
@@ -1783,6 +1791,8 @@ subroutine gsi_fv3ncdf2d_read(fv3filenamegin,it,ges_z,ges_t2m,ges_q2m)
 ! abstract: read in 2d fields from fv3_sfcdata file in mype_2d 
 !                Scatter the field to each PE 
 ! program history log:
+!   2023-02-14  Hu   -  Bug fix for read in subdomain surface restart files
+!
 !   input argument list:
 !     it    - time index for 2d fields
 !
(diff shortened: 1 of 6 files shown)


src/gsi/setupdbz.f90
```diff
@@ -1779,6 +1781,7 @@ subroutine init_netcdf_diag_
      end if
 
      call nc_diag_init(diag_conv_file, append=append_diag)
+     call flush(6)
```
TingLei-NOAA: Those "call flush". lines remaining from previous coding/debugging process could be removed

src/gsi/gsi_rfv3io_mod.f90
```diff
@@ -1902,40 +1912,24 @@ subroutine gsi_fv3ncdf2d_read(fv3filenamegin,it,ges_z,ges_t2m,ges_q2m)
...
-                   allocate(sfc(dim(dim_id(1)),dim(dim_id(2)),dim(dim_id(3))))
-                   iret=nf90_get_var(gfile_loc,i,sfc)
-                else if (ndim == 2) then
-                   allocate(sfc(dim(dim_id(1)),dim(dim_id(2)),1))
-                   iret=nf90_get_var(gfile_loc,i,sfc(:,:,1))
-                endif
-             else
-                write(*,*) "Mismatch dimension in surfacei reading:",nx,ny,dim(dim_id(1)),dim(dim_id(2))
-                call stop2(119)
+             allocate(sfc(nx,ny,1))
```
TingLei-NOAA, line 1928: A little confusing to me.  When ndim==3, in the original code, the sfc be as original 1930, now, it is always as the current line 1928, with the 3th dimension of 1.  What is supposed to happen when ndim ==3 , when sfc in the files are of vertical dimension >1 ?

jswhit:
The mods to enkf.f90 seem to be limiting the increments to maintain positive-definite dbz analysis values.  This is similar to what the 'cliptracers' logical namelist variable does.  Could that not be used instead?

Also, these changes also only apply to the serial filter.  Should they not also be applied to the LETKF?

chunhuazhou:
> The mods to enkf.f90 seem to be limiting the increments to maintain positive-definite dbz analysis values. This is similar to what the 'cliptracers' logical namelist variable does. Could that not be used instead?
> 
> Also, these changes also only apply to the serial filter. Should they not also be applied to the LETKF?

Hi Jeff, here is the comments I got from @Wangy1111 , the author of these changes. 
"The 'cliptracers' option is used to remove negative hydrometeors after the serial assimilation and the added changes aim to remove negative reflectivity during the serial assimilation. We find that spurious increments are produced due to the negative reflectivity during serial assimilation.

For the LETKF part, we don't use it for the assimilation. I am not sure if this method has the same problem as the EnSRF method."
Please let us know what you think. Thank you for your review!

jswhit:
I think cliptracers could be used for this, and then it would work for both serial and LETKF solvers.

Wangy1111:
> I think cliptracers could be used for this, and then it would work for both serial and LETKF solvers.

We notice that the 'cliptracers' option is used after the assimilation of all observations. Our changes are used in the serial assimilation of each observation.

jswhit:
> > I think cliptracers could be used for this, and then it would work for both serial and LETKF solvers.
> 
> We notice that the 'cliptracers' option is used after the assimilation of all observations. Our changes are used in the serial assimilation of each observation.

Is there a reason that enforcing positive-definiteness of dbz after each ob is assimilated is desirable?   It's not done for other positive-definite fields (like humidity).

Wangy1111:
> > > I think cliptracers could be used for this, and then it would work for both serial and LETKF solvers.
> > 
> > 
> > We notice that the 'cliptracers' option is used after the assimilation of all observations. Our changes are used in the serial assimilation of each observation.
> 
> Is there a reason that enforcing positive-definiteness of dbz after each ob is assimilated is desirable? It's not done for other positive-definite fields (like humidity).

The problem we found was due to inconsistencies with thresholding dbz before and during DA, and this caused some unstable amplifying behaviors near edges of strong precipitation when clear air and large reflectivity values are assimilated in locations near each other (as may be the case in the leading line of an MCS).

In our setup, clear air observations are set to 0 dbz, and we similarly threshold the first guess reflectivity so values can only be 0 and higher ensuring clear air ob innovations are <= 0. However, no such thresholding takes place during DA. Because we are using a serial filter, the background gets routinely updated for each ob assimilated and then used as a new background for the next obs' assimilation. To be consistent with our definitions (that 0 is the "clear air" value), then we also need to threshold each intermediate update of the background during the EnKF update. Otherwise, intermediate updates may cause values of the background to go below zero, which may cause spurious positive observation innovations even for clear air obs. This effect can get amplified near the edges of heavy precipitation because the observation innovation magnitudes in these regions may be very large. This is further true for the first reflectivity analysis, where the background and observations are most different from each other (again increasing observation innovations near precip regions)

We only saw the problem when assimilating radar reflectivity (not for other observations), probably due to its high data density.

Hope the above can explain the problem associated with our changes to enkf.f90. Thank you.

hu5970:
@jswhit Based on Yongming's explanation, I guess this problem should be there for any positive-definiteness background field that have sharp boundary from 0 to positive observations. Does EnKF analyze such kind of observations other than dBZ?

jswhit:
@Wangy1111 thanks for the explanation.  Could you add some comments in the code explaining this, so we remember why it is there?  (if there's a reference that would be great - does DART do this?).  I also wonder if this problem occurs with the LETKF - have you tried assimilating DBZ with the LETKF?

src/gsi/gsi_rfv3io_mod.f90
hu5970: The one version of FV3 model write out surface fields as 3 dimension fields. The third dimension of this array is always 1 but it fit to the requirement of the dimension number when nf90_get_var is used to read 3D or 2D array.

TingLei-daprediction:
An update on letkf using dbz using this PR: 
A preliminary test was run by simply setting letkf_flag=.true. for a rrfs enkf run using dbz. 
Some results at https://docs.google.com/presentation/d/182EEkdyMKhTwCj154dU6QvJkU4hQM7y9/edit?usp=sharing&ouid=104470932239919354212&rtpof=true&sd=true. 
I concluded for regional runs, this function with this PR using dbz as observations also works for letkf, though further tests/evaluation of the latter (the general regional letkf) would benefit a lot.

ShunLiu-NOAA:
@hu5970 Is it safe to proceed with this PR based on Ting's test results?

hu5970:
@chunhuazhou Could you sync with the develop branch and then we start to run regression test suite on Hera, Orion, and WCOSS2? Thanks, Ming

chunhuazhou:
> @chunhuazhou Could you sync with the develop branch and then we start to run regression test suite on Hera, Orion, and WCOSS2? Thanks, Ming

Done merging with the develop branch. Thank you, Ming!

jswhit:
@TingLei-NOAA thanks for checking dbz assimilation for the LETKF.   This PR should only affect the serial filter, so that the fact that appears to work fine in the LETKF is good news.  However, it does make me wonder if these changes to serial filter are really necessary.  I'm OK with merging anyway though.

TingLei-daprediction:
@jswhit  Thanks.  Sure.  It is very interesting for me to test/evaluate letkf for regional app. in the near future (and learn more from other groups who already has experiences  on this ).

chunhuazhou:
@jswhit @TingLei-daprediction Thank you very much for the review and testing the LETKF!

All the regression tests passed on Hera, as shown in the message below:
```

    Start 1: [=[global_3dvar]=]
    Start 2: [=[global_4dvar]=]
    Start 3: [=[global_4denvar]=]
    Start 4: [=[hwrf_nmm_d2]=]
    Start 5: [=[hwrf_nmm_d3]=]
    Start 6: [=[rtma]=]
    Start 7: [=[rrfs_3denvar_glbens]=]
    Start 8: [=[netcdf_fv3_regional]=]
    Start 9: [=[global_enkf]=]
1/9 Test #8: [=[netcdf_fv3_regional]=] ........   Passed  485.03 sec
2/9 Test #7: [=[rrfs_3denvar_glbens]=] ........   Passed  488.64 sec
3/9 Test #5: [=[hwrf_nmm_d3]=] ................   Passed  499.56 sec
4/9 Test #4: [=[hwrf_nmm_d2]=] ................   Passed  549.52 sec
5/9 Test #9: [=[global_enkf]=] ................   Passed  839.07 sec
6/9 Test #6: [=[rtma]=] .......................   Passed  1213.53 sec
7/9 Test #3: [=[global_4denvar]=] .............   Passed  1671.39 sec
8/9 Test #2: [=[global_4dvar]=] ...............   Passed  1683.81 sec
9/9 Test #1: [=[global_3dvar]=] ...............   Passed  1791.04 sec

100% tests passed, 0 tests failed out of 9

Total Test time (real) = 1791.05 sec 
```

hu5970:
Here are regression test results on Orion:
```
    Start 1: [=[global_3dvar]=]
    Start 2: [=[global_4dvar]=]
    Start 3: [=[global_4denvar]=]
    Start 4: [=[hwrf_nmm_d2]=]
    Start 5: [=[hwrf_nmm_d3]=]
    Start 6: [=[rtma]=]
    Start 7: [=[rrfs_3denvar_glbens]=]
    Start 8: [=[netcdf_fv3_regional]=]
    Start 9: [=[global_enkf]=]
1/9 Test #7: [=[rrfs_3denvar_glbens]=] ........***Failed  608.09 sec
2/9 Test #9: [=[global_enkf]=] ................   Passed  609.79 sec
3/9 Test #4: [=[hwrf_nmm_d2]=] ................   Passed  727.85 sec
4/9 Test #8: [=[netcdf_fv3_regional]=] ........   Passed  783.46 sec
5/9 Test #5: [=[hwrf_nmm_d3]=] ................***Failed  797.18 sec
6/9 Test #6: [=[rtma]=] .......................***Failed  1331.21 sec
7/9 Test #3: [=[global_4denvar]=] .............***Failed  1623.34 sec
8/9 Test #2: [=[global_4dvar]=] ...............   Passed  1922.63 sec
9/9 Test #1: [=[global_3dvar]=] ...............   Passed  1983.43 sec
```
Three cases,  rrfs_3denvar_glbens, rtma, and global_4denvar, have failed "the scalability test".
The hwrf_nmm_d3 failed in "The slope check"

No major failures from the regression tests on Orion. We can consider the regression tests passed on Orion.

TingLei-NOAA (approved):

shoyokota:
There are no concerns also for me now. I confirmed that the EnKF cycle of RRFS succeeded even with this GSI develop branch after this update and the result was completely the same as [the branch of the current RRFS](https://github.com/NOAA-GSL/GSI/tree/feature/rrfs_dev). I approve this PR.

shoyokota (approved):

TingLei-NOAA:
An update:  on wcoss2,  all tests passed!

hu5970:
@TingLei-daprediction  Thanks for your help with the tests.

hu5970 (approved):

PR #451, opened 2022-08-05
  GitHub Issue NOAA-EMC/GSI#450 Using global and regional ensembles simultaneously for fv3-lam GSI EnVar (add l_both_fv3sar_gfs_ens option)

TingLei-daprediction:
This PR is to add the mixed ensemble option for fv3-lam GSI EnVar.  https://github.com/NOAA-EMC/GSI/issues/450.
It passed the developer's own verification tests for this branch on hera.  On hera,  It failed in a few GSI 's own regression tests for the scalability problem, which I think could be attributed to some hera's system performance change and could be disregard now. I am re-running those regression tests and would update here.  
@MichaelLueken-NOAA  I did the rebase for this change and please let me know if it doesn't meet the GSI PR requirement.

PR #541, opened 2023-02-21
GitHub Issue NOAA-EMC/GSI#538 Add options to tune weights of mixed ensemble for FV3-LAM EnVar

### LINKED ISSUE #538: Add options to tune weights of mixed ensemble for FV3-LAM EnVar
In https://github.com/NOAA-EMC/GSI/pull/451, mixed ensemble (using global and regional ensembles simultaneously) options were introduced. However, weight of each ensemble was fixed to one, so background error covariances created by these summation were overestimated compared to only using global or regional ensembles. Here, the following options are introduced to tune weights of mixed ensembles. Setting in weight_ens_gfs+weight_ens_fv3sar=1 is preferable, but not necessary.

- weight_ens_gfs: weight of global ensemble (default: 1.0)
- weight_ens_fv3sar: weight of regional (FV3-LAM) ensemble (default: 1.0)

shoyokota:
This PR adds options (weight_ens_gfs and weight_ens_fv3sar) to tune weights of mixed ensemble for FV3-LAM EnVar (https://github.com/NOAA-EMC/GSI/issues/538). Regression tests for global 3dvar/4denvar/4dvar are not completed yet, but for other tests, issues are not found except for "failed the scalability test" and "exceeded maximum allowable hardware memory limit" on Orion.

Fixes #538

### DIFF
--- a/src/gsi/gsimod.F90
+++ b/src/gsi/gsimod.F90
@@ -154,7 +154,7 @@
                          i_ensloccov4tim,i_ensloccov4var,i_ensloccov4scl,l_timloc_opt,&
                          vdl_scale,vloc_varlist,&
                          global_spectral_filter_sd,assign_vdl_nml,parallelization_over_ensmembers
-  use hybrid_ensemble_parameters,only : l_both_fv3sar_gfs_ens,n_ens_gfs,n_ens_fv3sar
+  use hybrid_ensemble_parameters,only : l_both_fv3sar_gfs_ens,n_ens_gfs,n_ens_fv3sar,weight_ens_gfs,weight_ens_fv3sar
   use rapidrefresh_cldsurf_mod, only: init_rapidrefresh_cldsurf, &
                             dfi_radar_latent_heat_time_period,metar_impact_radius,&
                             metar_impact_radius_lowcloud,l_gsd_terrain_match_surftobs, &
@@ -1408,7 +1408,8 @@
 !                 s_ens_h  = v1L1  v2L1  v3L1  v1L2 v2L2
 !                 Then localization lengths will be assigned as above.
 !
-  namelist/hybrid_ensemble/l_hyb_ens,uv_hyb_ens,q_hyb_ens,aniso_a_en,generate_ens,n_ens,l_both_fv3sar_gfs_ens,n_ens_gfs,n_ens_fv3sar,nlon_ens,nlat_ens,jcap_ens,&
+  namelist/hybrid_ensemble/l_hyb_ens,uv_hyb_ens,q_hyb_ens,aniso_a_en,generate_ens,n_ens,&
+                l_both_fv3sar_gfs_ens,n_ens_gfs,n_ens_fv3sar,weight_ens_gfs,weight_ens_fv3sar,nlon_ens,nlat_ens,jcap_ens,&
                 pseudo_hybens,merge_two_grid_ensperts,regional_ensemble_option,fv3sar_bg_opt,fv3sar_ensemble_opt,full_ensemble,pwgtflg,&
                 jcap_ens_test,beta_s0,beta_e0,s_ens_h,s_ens_v,readin_localization,eqspace_ensgrid,readin_beta,&
                 grid_ratio_ens, &
@@ -1821,7 +1822,8 @@
     else 
        write(6,*)'n_ens_gfs and n_ens_fv3sar won"t be used if not regional_ensemble_option==5' 
     endif
-    
+    weight_ens_gfs=one
+    weight_ens_fv3sar=one
   endif
   if(ltlint) then
      if(vqc .or. njqc .or. nvqc)then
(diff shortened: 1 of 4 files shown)


shoyokota:
Is it possible to ask @TingLei-NOAA and @lbi2018 to review this PR? I think they are well-qualified because they are major contributers of #451.

src/gsi/gsimod.F90
```diff
@@ -1821,7 +1822,8 @@ subroutine gsimain_initialize
     else 
        write(6,*)'n_ens_gfs and n_ens_fv3sar won"t be used if not regional_ensemble_option==5' 
     endif
-    
+    weight_ens_gfs=one
```
TingLei-NOAA, line 1825: Here, the namelist parameters like l_both_fv3sar_gfs_ens had been readin as well as weight_ens_gfs and weight_ens_fv3sar, right?  then those assignments of the later two parameters (weight_ens??)  will override the readin values for them. So, they will always be one. 
Do I misunderstand something here?

shoyokota: In the case of l_both_fv3sar_gfs_ens=.false., only either global or regional ensemble is used, so weight_ens_gfs and weight_ens_fv3sar should be one (only either is used). To force that, I added this part. (If ensemble B should be tuned in this case, please don't set weight_ens_gfs or weight_ens_fv3sar but set beta_e0.) In the case of l_both_fv3sar_gfs_ens=.true., weight_ens_gfs and weight_ens_fv3sar are not overwritten and the parameters in the namelist are used.

TingLei-NOAA: @shoyokota Thanks for your explanation . That does make sense. 
This question is resolved

TingLei-NOAA (approved):

lbi2018 (approved):
Approved.

src/gsi/gsimod.F90
ShunLiu-NOAA: @shoyokota Could you please run a regression test and document the test result here.

shoyokota: @ShunLiu-NOAA I tried regression tests on Orion, but global 3dvar/4denvar/4dvar tests were failed because accesses to some files belonging to group "rstprod" are not permitted. I requested access to the NOAA RDHPCS Program, but it has not been accepted yet after two weeks. Even after it is accepted, I can run tests only on Orion. Could I ask someone to run the regression tests on my behalf?

ShunLiu-NOAA: @TingLei-NOAA Could you please run a regression test on HERA or WCOSS2?

TingLei-NOAA: @ShunLiu-NOAA  Sure, I will give an update here when I have finished the regression tests.

TingLei-NOAA:
@ShunLiu-NOAA   On hera,  two failed tests:  2:global_4dvar;7:rrfs_3denvar_glbens.   Seems GSI/fix/anavinfo.rrfs  is missing for 7 while GSI/fix/global_anavinfo_qlqi.l127.txt is  missing for 2  in this PR.  It seems I still use an old fix dir.  But "git submodule update" gave the error with message
"error: Server does not allow request for unadvertised object 0be26971f834fe9b1d5b118e1e0ffed53facf671
Fetched in submodule path 'fix', but it did not contain 0be26971f834fe9b1d5b118e1e0ffed53facf671. Direct fetching of that commit failed"   
Are anything  that I  missed for the current GSI downloading?

TingLei-NOAA:
@ShunLiu-NOAA ,  seems I should first run "git submodule sync" .   After that, "fix " dir was updated .  The regressiong tests of those two failed ones  is ongoing.

ShunLiu-NOAA:
> 
 I don't have any regression test on Hera recently. @hu5970 Could you work with Ting on this regression test?

TingLei-NOAA:
@ShunLiu-NOAA An update on my regression tests.   Thanks to Russ,  the current GSI regression tests procedure works fine for me on hera.  But a few re-run of the ctest always gave me a couple of failed tests (different for different runs) for their wall clock time exceeding the time limits (even I changed all of them to 30:00).  I will try more required time while I am wondering what the culprit is .

RussTreadon-NOAA:
@TingLei-NOAA , the most important aspect of the ctests is reproducibility independent of task and thread count.   A few of the checks in the current suite of ctests need to be revised.   While the test indicates _Failed_ it is not a fatal fail.   For example, slow I/O on Orion can lead to an anomalously large wall time.   This is a system fileset issue, not a gsi or enkf code issue.

TingLei-NOAA:
@RussTreadon-NOAA  Thank you for your clarification/reminding.  Yes. the job killed for exceeding time would cause the regression test become idle.   Since, such tests with this issue were different for different runs of the whole ctest (they succeeded in different runs), I would conclude that the current PR passed the current regression tests . @ShunLiu-NOAA .

ShunLiu-NOAA (approved):

PR #559, opened 2023-04-13
Fix write_incr

### LINKED ISSUE #558: Issues with write_incr
While trying to run the gsi in a different configuration, my runs would not complete properly (both the control and test).  After spending some time trying to figure out what I did wrong, I eventually concluded that the problem was in the code.

Modifications to the code were made in my converge4/GSI/fix_write_incr directory to improve the routine.

The primary change was to add 
use gsi_bundlemod, only: assignment(=)

This allows the code to handle the bundlea =bundleb properly and the code then completes.

Also, the allocate_state and allocate_pred routines were reordered and deallocates were added.  This is to eliminate the memory leak.

All regression test results were the same and case I was using complete properly.

jderber-NOAA:
Fixes #558


src/gsi/write_incr.f90
```diff
@@ -202,7 +211,7 @@ subroutine write_fv3_inc_ (grd,sp_a,filename,mype_out,gfs_bundle,ibin)
     iqg = getindex(svars3d,'qg')
 
     istatus=0
-    call gsi_bundlegetpointer(gfs_bundle,'q', sub_qanl, iret); istatus=istatus+iret
+    call gsi_bundlegetpointer(svalinc(ibin2),'q', sub_qanl, iret); istatus=istatus+iret
```
CoryMartin-NOAA: good catch on this, no idea why this one is different from the rest... does this have any implcations for current GFSv16?

CoryMartin-NOAA (dismissed):
Did not test but changes look straightforward and correct to me

src/gsi/write_incr.f90
TingLei-NOAA: @jderber-NOAA   I am a little confused here.  In the following around line 313 (in this program), there is 
`  ! compute q increment
    sub_q = sub_qanl(:,:,:) - ges_q1(:,:,:,ibin)
`
So, seems to me, the original code intentionally gets analysis q in full at the above (line 205) .   
I am wondering why the original treatment is used .  However, if the change of line 214 is used, the following line 313 should also be changed too?

jderber-NOAA: I agree good catch.  I do not understand why this variable is handled differently.  But the intent of this change is not to change the science, but to make the code work.  The line is reverted to use gfs_bundle and the tests resubmitted.

RussTreadon-NOAA:
As per the 4/10/2023 email sent to the GSI subversion email list, PRs must be merged within 6 weeks of creation.  PRs not merged within the 6 week window will be returned to the originating developer (see [GSI Wiki](https://github.com/NOAA-EMC/GSI/wiki/GSI---Code-Management-Policy)).   

This PR was opened 4/13/2023.   The due date for this PR is set to **5/25/2023**, six weeks after 4/13/2023.

TingLei-NOAA (approved):

CoryMartin-NOAA (approved):

src/gsi/write_incr.f90
jderber-NOAA: All regression tests passed and test of code was fine.  I believe this is ready for re-review.

RussTreadon-NOAA:
All regression tests pass.  Two peer reviews approve.   I'll reach out to the Handling Review Team for final approval and merge.

RussTreadon-NOAA (approved):
Approve.

jderber-NOAA:
Thanks Cory and Ting!  Up to Russ now!

jderber-NOAA:
I seem to be a few minutes (or less) behind everyone this morning!

TingLei-NOAA:
@jderber-NOAA  A following up question ,  the previous assignment between gsi bundle (like currently line 195) while without defining the operator overload by  "use gsi_bundlemod, only: assignment(=)"  added in this PR,   would it cause any actual impacts?  Namely, will this PR create different increment (correct) fields in the generated increment files?  I am not sure if current regression tests compare the generated increment files.

jderber-NOAA:
Ting, That is a good question.  I don't know.  For my test case, the code would not complete without these changes so I could not compare.  Could you do a comparison?  If there is a difference, I am certain that the new is the more correct one.

TingLei-NOAA:
@jderber-NOAA   I am not sure if current regression tests for global gsi will create the inc files. 
I will talk to global DA developers off line for help on this.  I will give an update as soon as possible.

RussTreadon-NOAA:
Merger into `develop` on hold pending feedback from @TingLei-NOAA

TingLei-NOAA:
An update:  Thanks to @RussTreadon-NOAA 's  information,  I directly compared the increment files in the regression test : global_3dvar on hera:  in /scratch1/NCEPDEV/stmp2/Ting.Lei/GSI/tmpreg_global_3dvar, I ran 
^
 >> diff -h global_3dvar_hiproc_contrl/siginc.nc global_3dvar_hiproc_updat/siginc.nc
V
There are none of output.  So, I conclude this means they are identical . 
I am having some issues on dogwood (the wcoss2 machine available to me now) for git . So I haven't run tests on it.

RussTreadon-NOAA:
Given @TingLei-NOAA confirmation of identical increments go ahead and merge this PR into `develop`

PR #567, opened 2023-05-04
For EMC GSI issue#566: Feature/enkf q clipping: adding the clipping of negative values of moisture in FV3-LAM EnKF

TingLei-daprediction:
In correspondence to  the EMC GSI Issue#566, this PR contains a quick adding of the clipping of negative values of sphum (q) in the analysis of FV3-LAM EnKF. 
This part of codes are not tested in the current GSI regression tests, which, hence, are not run. 
The current codes are verified using local FV3-LAM case.  It is found the differences from this changes exist for sphum ( maximum values about 0.003 (units) and 0.3 K for T (sensible T).   The latter is because the sphum would be used when the analysis TV is converted to T.  All differences are  on spontaneous points and values are reasonable as expected. 
Hence, the code is regarded as verified.


TingLei-daprediction:
The changes for build.sh was added to this PR accidently. It is reversed to the original one from the develop branch of the fork and still was displayed in the changed files. Sorry for this unnecessary item.

ush/build.sh
```diff
@@ -12,7 +12,7 @@ COMPILER=${COMPILER:-"intel"}
 BUILD_DIR=${BUILD_DIR:-"${DIR_ROOT}/build"}
 INSTALL_PREFIX=${INSTALL_PREFIX:-"${DIR_ROOT}/install"}
 GSI_MODE=${GSI_MODE:-"Regional"}  # By default build Regional GSI (for regression testing)
-ENKF_MODE=${ENKF_MODE:-"GFS"}     # By default build Global EnKF  (for regression testing)
+ ENKF_MODE=${ENKF_MODE:-"GFS"}     # By default build Global EnKF  (for regression testing)
```
RussTreadon-NOAA, line 15: I don't think this extra white space change is meant to be part of this PR.

TingLei-daprediction: @RussTreadon-NOAA  please see the above explanation.
If a cleaner PR is needed, I will try to reverse it.

RussTreadon-NOAA: Yes, please revert this change.  `ush/build.sh` should not be part of this PR.   

The simplest thing to do is to remove the white space from `ush/build.sh` in a local working copy of your branch, commit this change, and push the local copy of your branch to github.   Then we should not see `ush/build.sh` in the list of files changed.

chunhuazhou (approved):
Looks good to me.

TingLei-daprediction:
Dear all, @ShunLiu-NOAA @chunhuazhou @daviddowellNOAA  Sorry I find the easiest way for a cleaner PR is to open another PR with a new branch 
https://github.com/NOAA-EMC/GSI/pull/568

ush/build.sh
TingLei-daprediction: @RussTreadon-NOAA   Thanks for your suggestions. Seems some tricky issues occured and I found the easier to open another PR with a different branch.

PR #576, opened 2023-05-18
GitHub Issue NOAA-EMC/GSI# 572 OU MAP lab developments of convective-…

### LINKED ISSUE #572: OU MAP lab developments of convective-scale static background error covariance within GSI for FV3-LAM
A convective-scale static background error covariance (**B**) is developed by OU MAP lab for direct assimilation of radar reflectivity within GSI-based hybrid EnVar system (Wang and Wang 2021, MWR). Such static **B** is further developed for FV3-LAM by OU MAP lab (Yang et al., 2023).
 
The developments include,
(1) extension of control variables (CVs) to include hydrometeors, vertical velocity, and reflectivity.
(2) Utilization of (u, v) and rh as the optimal momentum and humidity CVs.
(3) reflectivity-dependent binning method for static covariance statistics.
(4) adaptive incorporation of static **B** in hybrid EnVar system based on the quality of ensemble covariances.
(5) extension of balance operator to include cross-variable correlations that are suitable for convective scales.

YueYang-OU:
**Description**

The PR adds the following capabilities of the convective-scale static BEC developed by OU MAP lab,
(1) extension of control variables (CVs) to include hydrometeors, vertical velocity, and reflectivity.
(2) Utilization of (u, v) and rh as the optimal momentum and humidity CVs.
(3) reflectivity-dependent binning method for static covariance statistics.
(4) adaptive incorporation of static B in hybrid EnVar system based on the quality of ensemble covariances.
(5) extension of balance operator to include cross-variable correlations that are suitable for convective scales.
 
Fixes #572

**Type of change**

- [X] New feature (non-breaking change which adds functionality)
  
**Checklist**

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] New and existing tests pass with my changes

**DUE DATE** for merger of this PR into develop is **6/29/2023** (six weeks after PR creation).


shoyokota:
@YueYang-OU I'm trying testing your code with 'if_cs_staticB=.true.', but it failed in inputting variables 'mask' and 'maskh'. What are these variables? I guess 'mask' is used for reflectivity-dependent binning and 'maskh' is used for adaptive incorporation of static B, but how should we create and input these variables? I would like to know also how to set 'anavinfo' for using this option.

YueYang-OU:
> @YueYang-OU I'm trying testing your code with 'if_cs_staticB=.true.', but it failed in inputting variables 'mask' and 'maskh'. What are these variables? I guess 'mask' is used for reflectivity-dependent binning and 'maskh' is used for adaptive incorporation of static B, but how should we create and input these variables? I would like to know also how to set 'anavinfo' for using this option.

@shoyokota Yes, the variable 'mask' is the geographical mask for the reflectivity-dependent binning, and 'maskh' is the geographical mask for the consistency ratio (CR)-based adaptive hybridization. These two variables are input via the 'fv3_mask' file that comes from the 'fv3_tracer' file. Specifically, we copy 'fv3_tracer' into 'fv3_mask', and then we change the values of 'rainwat' as 'mask' and of 'graupel' as 'maskh'.

In terms of the diagnostics of 'mask', the static **B** matrix from the bin of strong storm is applied on the locations where the observed composite reflectivity exceeds (>=) 35 dBZ, and the static **B** matrix from the bin of weak storm is applied on the locations where the observed composite reflectivity between 10 and 35 dBZ. For the definition of 'maskh', the static **B** matrix with a fixed covariance weight is applied on the locations with CR less than a value (between 0 and 1). The CR denotes the ratio of the total spread (including the sqrt of the sum of observation error variance and ensemble variance) over the square root of the innovation (observation-minus-background) for reflectivity. Please refer to Wang and Wang (2021) for more details. 

Here is the [anavinfo](https://drive.google.com/file/d/1-eCixQw6TigyZ1HZeUTtbWIk_7QVZxZA/view?usp=sharing) we use.

shoyokota:
@YueYang-OU Thank you for clarifying 'mask' and 'maskh'.

> @shoyokota Yes, the variable 'mask' is the geographical mask for the reflectivity-dependent binning, and 'maskh' is the geographical mask for the consistency ratio (CR)-based adaptive hybridization. These two variables are input via the 'fv3_mask' file that comes from the 'fv3_tracer' file. Specifically, we copy 'fv3_tracer' into 'fv3_mask', and then we change the values of 'rainwat' as 'mask' and of 'graupel' as 'maskh'.

Does it mean that we should create 'fv3_mask' based on dbz and CR before running GSI? If so, could you give us the sample code to make 'mask' and 'maskh'?

If possible, I think it is better to create and use 'mask' and 'maskh' in GSI without additional input these data from 'fv3_mask'. (I believe it is possible.) Why didn't you apply such implementation?

Even in the case to input these data, I think 'fv3_mask' should have only 'mask' and 'maskh' data if GSI input only them from 'fv3_mask'. Why should we copy 'fv3_tracer' into 'fv3_mask'?

> Here is the [anavinfo](https://drive.google.com/file/d/1-eCixQw6TigyZ1HZeUTtbWIk_7QVZxZA/view?usp=sharing) we use.

I cannot access the anavinfo. Could you permit my access?

wangym1111:
> @YueYang-OU Thank you for clarifying 'mask' and 'maskh'.
> 
> > @shoyokota Yes, the variable 'mask' is the geographical mask for the reflectivity-dependent binning, and 'maskh' is the geographical mask for the consistency ratio (CR)-based adaptive hybridization. These two variables are input via the 'fv3_mask' file that comes from the 'fv3_tracer' file. Specifically, we copy 'fv3_tracer' into 'fv3_mask', and then we change the values of 'rainwat' as 'mask' and of 'graupel' as 'maskh'.

> 
> Does it mean that we should create 'fv3_mask' based on dbz and CR before running GSI? If so, could you give us the sample code to make 'mask' and 'maskh'?

We rely on the GSI code and make changes to it to create 'mask' and 'maskh'. You can follow this branch to see how to create them.  https://github.com/wangyongm/GSI_staticB_mask.git

> 
> If possible, I think it is better to create and use 'mask' and 'maskh' in GSI without additional input these data from 'fv3_mask'. (I believe it is possible.) Why didn't you apply such implementation?

Yes, it is possible to implement the option. Note that the above approach in @YueYang-OU's comment for creating 'mask' and 'maskh' is only a potential one. We also have several other ideas (e.g., machine learning-based) to create them. Therefore, we keep the generation of fv3_mask outside GSI for more flexibility in the future.
> 
> Even in the case to input these data, I think 'fv3_mask' should have only 'mask' and 'maskh' data if GSI input only them from 'fv3_mask'. Why should we copy 'fv3_tracer' into 'fv3_mask'?

As mentioned above, we rely on GSI code to create fv3_mask, thus all fields are required here.

> 
> > Here is the [anavinfo](https://drive.google.com/file/d/1-eCixQw6TigyZ1HZeUTtbWIk_7QVZxZA/view?usp=sharing) we use.
> 
> I cannot access the anavinfo. Could you permit my access?

Yue has reset the permission and you can try it again.

Thank you for your comments.

shoyokota:
@wangym1111 Thank you for the explanation.

> We rely on the GSI code and make changes to it to create 'mask' and 'maskh'. You can follow this branch to see how to create them. https://github.com/wangyongm/GSI_staticB_mask.git

I looked for 'mask' and 'maskh' in your branch, but I couldn't find them. Where should I see?

> Yes, it is possible to implement the option. Note that the above approach in @YueYang-OU's comment for creating 'mask' and 'maskh' is only a potential one. We also have several other ideas (e.g., machine learning-based) to create them. Therefore, we keep the generation of fv3_mask outside GSI for more flexibility in the future.

I understand your thought, but as you explained, creating 'fv3_mask' is relied on GSI code (despite you create it outside GSI), right? If so, I think it is better to add the option to create 'fv3_mask' with the current way in the GSI code at least for now (even if we consider flexibility in the future). Actually so far, it was difficult for us to run tests of the convective static B using your current code that demands to create 'fv3_mask' by ourselves. Could you add the option to create 'fv3_mask' also in this GSI?

> As mentioned above, we rely on GSI code to create fv3_mask, thus all fields are required here.

I'm still confusing. I asked why many fields of 'fv3_mask' are duplicated with 'fv3_tracer'. Since we can use 'fv3_tracer', it is not required to input/output duplicated fields in 'fv3_mask'. I think redundant parts should be removed as much as possible in the authoritative GSI repository.

> Yue has reset the permission and you can try it again.

Thank you. I downloaded it.

wangym1111:
@shoyokota
We should clarify that the GSI code used to create fv3_mask is a separate one (GSI-mask) and different from this PR's GSI. 

GSI-mask is used to 1) convert reflectivity observations to model space and 2) diagnose values of 'mask' and 'maskh' following the definitions in @YueYang-OU 's comments.  Inside the code, the value of 'mask' is written to the variable 'rainwat', and the value of 'maskh' is written into the variable 'graupel'. Therefore, you need to run the GSI-mask individually to generate fv3_mask.  This also explains why the other fields are kept in fv3_mask.

We also have an internal discussion about merging the above processes into this PR's GSI. However, numerous additional modifications are required to setupdbz.f90, intdbz.f90, pcgsoi.f90, etc. It is not a trivial work. 

GSI-mask is only a temporary option and we definitely will provide a simpler approach to generate fv3_mask soon.

We can also share the GSI-mask code on Orion as well as the script using it in a separate email (this version can be applied to the fv3lam model directly) if you are still interested in this option.

Thank you.

shoyokota:
@wangym1111 

> GSI-mask is used to 1) convert reflectivity observations to model space and 2) diagnose values of 'mask' and 'maskh' following the definitions in @YueYang-OU 's comments. Inside the code, the value of 'mask' is written to the variable 'rainwat', and the value of 'maskh' is written into the variable 'graupel'. Therefore, you need to run the GSI-mask individually to generate fv3_mask. This also explains why the other fields are kept in fv3_mask.

I understand what is done by GSI-mask. It is a nice method as an option for temporally conducted experiments, but is redundant and easily causes confusion. It is better to modify this branch not to assume including the redundant fields in 'fv3_mask' before merging to the NOAA-EMC develop branch.

> We also have an internal discussion about merging the above processes into this PR's GSI. However, numerous additional modifications are required to setupdbz.f90, intdbz.f90, pcgsoi.f90, etc. It is not a trivial work.
> GSI-mask is only a temporary option and we definitely will provide a simpler approach to generate fv3_mask soon.

That sounds great! I understand that it is not necessarily trivial work, but I think it is better to wait for merging this branch until either is provided.

> We can also share the GSI-mask code on Orion as well as the script using it in a separate email (this version can be applied to the fv3lam model directly) if you are still interested in this option.

I'm still very interested in testing the convective static B even if merging this branch would be put off. It would be so helpful for the test if you could share the GSI-mask code on Orion. Could you let me use it?

Thanks a lot!

shoyokota:
@wangym1111 @YueYang-OU 
I tested the convective static B using your GSI-mask code, and I found some modifications were needed for the test. Could I ask following questions?

- To use the convective static B, control variables should be changed from (sf,vp) to (u,v), but in the calculation of ensemble B, there are some parts where only (sf,vp) is allowed (e.g., [cplr_get_fv3_regional_ensperts.f90#L455](https://github.com/YueYang-OU/GSI/blob/develop/src/gsi/cplr_get_fv3_regional_ensperts.f90#L455), [write_fv3_spread.f90#L141](https://github.com/YueYang-OU/GSI/blob/develop/src/gsi/write_fv3_spread.f90#L141)). So, my test failed in the setting of hybrid EnVar. Could you fix this issue?
- When sst is one of the control variables, the test failed probably because there is a part using hwll(i,1,:,nrf3_sf) for sst ([m_berror_stats_reg.f90#L897](https://github.com/YueYang-OU/GSI/blob/develop/src/gsi/m_berror_stats_reg.f90#L897)). How should we do to include sst in control variables?
- Some component of sigma_avn in the convective static B received from you looks zero. So, log(sigma_avn) in [m_berror_stats_reg.f90#L417](https://github.com/YueYang-OU/GSI/blob/develop/src/gsi/m_berror_stats_reg.f90#L417) couldn't calculated. Could you fix this issue? (If we insert "if(sigma_avn(k)<tiny_single) sigma_avn(k)=tiny_single", this issue may be resolved.)

src/gsi/prewgt_reg.f90
```diff
@@ -220,7 +231,13 @@ subroutine prewgt_reg(mype)
   end if
 
 ! Read in background error stats and interpolate in vertical to that specified in namelist
-  call berror_read_wgt_reg(msig,mlat,corz,corp,hwll,hwllp,vz,rlsig,varq,qoption,varcw,cwoption,mype,inerr)
+  if( if_cs_staticB )then
+     call berror_read_wgt_reg_extra(msig,mlat,num_bins2d,corz,corp,hwll,hwllp,vz,rlsig,varq,qoption,varcw,cwoption,mype,inerr)
+     print*,'hwll1',maxval( hwll(:,:,:,1) ), minval( hwll(:,:,:,1) )
+     print*,'hwll2',maxval( hwll(:,:,:,2) ), minval( hwll(:,:,:,2) )
```
shoyokota: The print to log file is not needed here.

src/gsi/prewgt_reg.f90
```diff
@@ -711,6 +758,11 @@ subroutine prewgt_reg(mype)
 
 ! Set up scales
 
+  if( if_cs_staticB )then
+     bin_l = 2 ! temporily assign the scale in second bin to all bins 
+  else
+     bin_l = 1
+  end if
```
shoyokota, line 768: This bin_l is not used. I think it can be deleted.

src/gsi/prewgt_reg.f90
```diff
@@ -725,7 +777,7 @@ subroutine prewgt_reg(mype)
            if (nn/=nrf3_oz) then
               if (k1 >= ks(k))then
                  l=int(rllat(ny/2,nx/2))
-                 fact=one/hwll(l,k1,nn)
+                 fact=one/hwll(l,k1,1,nn)
```
shoyokota: Here only hwll( : , : , 1 , : ) is used, so hwll( : , : , 2:3 , : ) is not used at all although it is input. Is this behavior as you expect?

src/gsi/prewgt_reg.f90
```diff
@@ -780,7 +832,7 @@ subroutine prewgt_reg(mype)
                     lp=min0(l+1,llmax)
                     dl2=rllat(j,i)-float(l)
                     dl1=one-dl2
-                    fact=cc/(dl1*hwllp(l,nn)+dl2*hwllp(lp,nn))
+                    fact=cc/(dl1*hwllp(l,1,nn)+dl2*hwllp(lp,1,nn))
```
shoyokota: Here only hwllp( : , 1 , : ) is used, so hwllp( : , 2:3 , : ) is not used at all although it is input. Is this behavior as you expect?

src/gsi/m_berror_stats_reg.f90
```diff
@@ -290,6 +326,801 @@ subroutine berror_read_bal_reg(msig,mlat,agvi,bvi,wgvi,mype,unit)
...
+     loc=nrf2_loc(n)
+     if (nrf_err(loc)) cycle
+     if (n==nrf2_sst) then
+        do i=1,mlat
+           corp(i,:,n)=zero_3
+        end do
+        do i=0,mlat+1
+           hwllp(i,:,n)=hwll(i,1,:,nrf3_sf)
+           hwllp(i,:,nc2d+1)=hwll(i,1,:,nrf3_sf) !not very nice, since it assumes that stl and sti
+           hwllp(i,:,nc2d+2)=hwll(i,1,:,nrf3_sf) !are always the first motley variables in convinfo
```
shoyokota: The convective-scale static B doesn't use sf, so my test failed here. Could you fix to be able to include sst to one of the control variables?

src/gsi/m_berror_stats_reg.f90
```diff
@@ -290,6 +326,801 @@ subroutine berror_read_bal_reg(msig,mlat,agvi,bvi,wgvi,mype,unit)
...
+  nrf2_tcamt=getindex(cvars2d,'tcamt')
+  nrf2_lcbas=getindex(cvars2d,'lcbas')
+  nrf2_cldch=getindex(cvars2d,'cldch')
+  nrf2_uwnd10m=getindex(cvars2d,'uwnd10m')
+  nrf2_vwnd10m=getindex(cvars2d,'vwnd10m')
+
+  if(nrf3_q>0 .and. qoption==2)then
+     do k=1,nsig
+        do j=1,mlat
+           varq(j,k)=min(max(real(corz(j,k,1,nrf3_q),r_kind),0.0015_r_kind),one)
```
shoyokota, line 789: For the case of qoption=2, I think varq also should be varq( 1:mlat , 1:nsig , 1:num_bins2d ) and binned by reflectivity.

src/gsi/m_berror_stats_reg.f90
```diff
@@ -290,6 +326,801 @@ subroutine berror_read_bal_reg(msig,mlat,agvi,bvi,wgvi,mype,unit)
...
+     hwll(:,:,:,nrf3_cw)=hwll(:,:,:,nrf3_q)
+     vz(:,:,:,nrf3_cw)=vz(:,:,:,nrf3_q)
+  end if
+  if ((.not. cwcoveqqcov_) .and. nrf3_cw>0) then
+     corz(:,:,:,nrf3_cw)=zero
+     if (cwoption==2) then
+        do k=1,nsig
+           if (ges_prslavg(k)>15.0_r_kind) then
+              do j=1,mlat
+                 varcw(j,k)=max(real(corz(j,k,1,nrf3_cw),r_kind),zero)
```
shoyokota, line 830: For the case of cwoption=2, I think varcw also should be varcw( 1:mlat , 1:nsig , 1:num_bins2d ) and binned by reflectivity.

src/gsi/m_berror_stats_reg.f90
```diff
@@ -290,6 +326,801 @@ subroutine berror_read_bal_reg(msig,mlat,agvi,bvi,wgvi,mype,unit)
...
+       read(inerr)agv_avn,bv_avn,wgv_avn ,extravi3
+    end if
+    close(inerr)
+
+!   compute vertical(pressure) interpolation index and weight
+    do k=1,nsig
+       rlsig(k)=log(ges_prslavg(k)/ges_psfcavg)
+    enddo
+    do k=1,msig
+       rlsigo(k)=log(sigma_avn(k))
```
shoyokota, line 411: My test failed here because some of sigma_avn(k) were zero. Could you add the treatment for the case of sigma_avn(k)=zero?

src/gsi/m_berror_stats_reg.f90
```diff
@@ -290,6 +326,801 @@ subroutine berror_read_bal_reg(msig,mlat,agvi,bvi,wgvi,mype,unit)
...
+           corp(i,:,n0+n)=corp(i,:,nrf2_vwnd10m)
+        end do
+        do i=0,mlat+1
+           hwllp(i,:,n0+n)=hwllp(i,:,nrf2_vwnd10m)
+        end do
+     endif
+  enddo
+
+  deallocate(nrf3_loc,nrf2_loc,nmotl_loc)
+
```
shoyokota, line 1120: I think normalization of vz (coordinate transformation from 1/(sigma) to 1/(grid unit) ) is required here as in the last part of subroutine berror_read_wgt_reg.

src/gsi/prewgt_reg.f90
```diff
@@ -455,11 +480,21 @@ subroutine prewgt_reg(mype)
...
+                       dssv(i,j,k,n)=(dl1*ozmzt(l,k)+dl2*ozmzt(l2,k))*dsv(1,k,1,llmin)
+                    end if
+                 else
+                    dssv(i,j,k,n)=(dl1*ozmzt(l,k)+dl2*ozmzt(l2,k))*dsv(1,k,1,llmin)
+                 end if
               end do
            end do
         end do
-     else if ( n==nrf3_qr .or. n==nrf3_qs .or. n==nrf3_qg .or. n==nrf3_qnr) then
+     else if ( l_be_T_dep .and. (n==nrf3_qr .or. n==nrf3_qs .or. n==nrf3_qg .or. n==nrf3_qnr) ) then
```
shoyokota: Why is l_be_T_dep added? If it is added here, the following 'if(l_be_T_dep) ... else ... endif' is not required.

src/gsi/balmod.f90
```diff
@@ -891,6 +990,261 @@ subroutine tbalance(t,p,st,vp,fpsproj,fut2ps)
...
+             end do
+          end if
+       end if
+
+    end do
+
+    return
+
+  end subroutine tbalance_extra
+
```
shoyokota, line 1245: I think these subroutines (det_field, balance_extra, and tbalance_extra) should be generalized to be able to apply for any number of nb_cv and any kinds of 3D and 2D variables.

src/gsi/bkerror.f90
```diff
@@ -94,8 +106,14 @@ subroutine bkerror(grady)
   call timer_ini('bkerror')
 
 ! Only need to get pointer for ii=1 - all other are the same
-  call gsi_bundlegetpointer ( grady%step(1), (/'t ','sf','vp','ps'/), &
-                              ipnts, istatus )
+  if( if_cs_staticB )then
+     call gsi_bundlegetpointer ( grady%step(1), (/'t ','u','v','ps'/), &
+                                 ipnts, istatus )
```
shoyokota, line 111: It is better to add 'q', 'w', 'qr', 'qs', 'qg', 'qi', 'ql', and 'dbz'.

src/gsi/directDA_radaruse_mod.f90
```diff
@@ -147,7 +147,7 @@ subroutine init_radaruse_directDA
     l_use_cvpqx         = .FALSE.           ! (not) use power transform to qx(qr/qs/qg) 
     cvpqx_pval          = 0.000001_r_kind   ! power transform value; nearly the same as use log transform to qx(qr/qs/qg) 
 
-    l_plt_be_stats      = .TRUE.            ! output BE stats for plot
+    l_plt_be_stats      = .FALSE.            ! output BE stats for plot
```
shoyokota: Why did you change the default l_plt_be_stats?

src/gsi/gsi_rfv3io_mod.f90
```diff
@@ -104,13 +107,15 @@ module gsi_rfv3io_mod
...
   character(len=max_varname_length), dimension(nphyvarslist), parameter :: &
     varphyvars = [character(len=max_varname_length) :: 'dbz']
-  character(len=max_varname_length), dimension(16+naero_cmaq_fv3+7+naero_smoke_fv3), parameter :: &
+  character(len=max_varname_length), dimension(nmskvarslist), parameter :: &
+    varmskvars = [character(len=max_varname_length) :: 'mask','maskh']
+  character(len=max_varname_length), dimension(16+naero_cmaq_fv3+7+naero_smoke_fv3+2), parameter :: &
     varfv3name = [character(len=max_varname_length) :: &
       'u','v','W','T','delp','sphum','o3mr','liq_wat','ice_wat','rainwat','snowwat','graupel','rain_nc','ref_f3d','ps','DZ', & 
-      aeronames_cmaq_fv3,'pm25at','pm25ac','pm25co','pm2_5','amassi','amassj','amassk',aeronames_smoke_fv3], &
+      aeronames_cmaq_fv3,'pm25at','pm25ac','pm25co','pm2_5','amassi','amassj','amassk',aeronames_smoke_fv3,'rainwat','graupel'], &
```
shoyokota, line 118: Currently, 'rainwat' and 'graupel' in fv3_mask are regarded as 'mask' and 'maskh', respectively, but these names cause confusion. I suggest to use other names after making the GSI-mask code more simply.

src/gsi/hybrid_ensemble_isotropic.F90
```diff
@@ -2685,6 +2693,8 @@ subroutine sqrt_beta_s_mult_cvec(grady)
      call stop2(999)
   endif
 
+  call gsi_bundlegetpointer (gsi_metguess_bundle(1),'maskh',ges_mask1,istatus)
```
shoyokota: It should be enclosed within 'if(  if_cs_staticB .and. if_consistency_ratio )'.

src/gsi/hybrid_ensemble_isotropic.F90
```diff
@@ -2783,6 +2806,8 @@ subroutine sqrt_beta_s_mult_bundle(grady)
...
```
shoyokota: It should be enclosed within 'if(  if_cs_staticB .and. if_consistency_ratio )'.

src/gsi/hybrid_ensemble_isotropic.F90
```diff
@@ -2858,11 +2895,15 @@ subroutine sqrt_beta_e_mult_cvec(grady)
...
   character(len=*),parameter::myname_=myname//'*sqrt_beta_e_mult'
-  integer(i_kind) :: i,j,k,ii,nn,ig
+  integer(i_kind) :: i,j,k,ii,nn,ig,istatus
+  real(r_kind),dimension(:,:,:),pointer::ges_mask1=>NULL()
+  real(r_kind)    :: sqrt_beta_e_temp
 
   ! Initialize timer
   call timer_ini('sqrt_beta_e_mult')
 
+  call gsi_bundlegetpointer (gsi_metguess_bundle(1),'maskh',ges_mask1,istatus)
```
shoyokota: It should be enclosed within 'if(  if_cs_staticB .and. if_consistency_ratio )'.

src/gsi/hybrid_ensemble_isotropic.F90
```diff
@@ -2925,19 +2978,32 @@ subroutine sqrt_beta_e_mult_bundle(aens)
...
   character(len=*),parameter::myname_=myname//'*sqrt_beta_e_mult'
-  integer(i_kind) :: i,j,k,nn,ig
+  integer(i_kind) :: i,j,k,nn,ig,istatus
+  real(r_kind),dimension(:,:,:),pointer::ges_mask1=>NULL()
+  real(r_kind)    :: sqrt_beta_e_temp
 
   ! Initialize timer
   call timer_ini('sqrt_beta_e_mult')
 
+  call gsi_bundlegetpointer (gsi_metguess_bundle(1),'maskh',ges_mask1,istatus)
```
shoyokota: It should be enclosed within 'if(  if_cs_staticB .and. if_consistency_ratio )'.

src/gsi/prewgt_reg.f90
```diff
@@ -431,15 +448,23 @@ subroutine prewgt_reg(mype)
      nrf2_loc(ii)=getindex(cvars,cvars2d(ii))
   enddo
 
+  nrf3_dbz =getindex(cvars3d,'dbz')
+  call gsi_bundlegetpointer(gsi_metguess_bundle(1),'mask',ges_mask,istatus)
```
shoyokota: It should be enclosed within 'if(  if_cs_staticB )'.

src/gsi/gsi_rfv3io_mod.f90
```diff
@@ -1357,6 +1414,17 @@ subroutine read_fv3_netcdf_guess(fv3filenamegin)
             if(if_model_dbz) &
                call GSI_BundleGetPointer ( GSI_MetGuess_Bundle(it), 'dbz' , ges_dbz ,istatus );ier=ier+istatus
          end if
+
+         if( if_cs_staticB )then
+            call GSI_BundleGetPointer ( GSI_MetGuess_Bundle(it), 'mask',ges_mask,istatus );ier=ier+istatus
+            if (ier/=0) call die(trim(myname),'cannot get pointers for met-fields, ier =',ier)
+            adaptive_hybrid = .true.
+         end if
+         if( adaptive_hybrid )then
```
shoyokota: What is difference between if_consistency_ratio and adaptive_hybrid? If they are almost the same, it is better to remove adaptive_hybrid and to change here to 'if(  if_cs_staticB .and. if_consistency_ratio )'.

src/gsi/prewgt_reg.f90
```diff
@@ -740,7 +792,7 @@ subroutine prewgt_reg(mype)
                        lp=min0(l+1,llmax)
                        dl2=rllat(j,i)-float(l)
                        dl1=one-dl2
-                       fact=one/(dl1*hwll(l,k1,nn)+dl2*hwll(lp,k1,nn))
+                       fact=one/(dl1*hwll(l,k1,1,nn)+dl2*hwll(lp,k1,1,nn))
```
shoyokota: Here only hwll( : , : , 1 , : ) is used, so hwll( : , : , 2:3 , : ) is not used at all although it is input. Is this behavior as you expect?

src/gsi/prewgt_reg.f90
```diff
@@ -654,7 +701,7 @@ subroutine prewgt_reg(mype)
      loc=nrf2_loc(n)
      do j=llmin,llmax
         do i=1,lon2
-           dsvs(i,j)  =corp(j,n)*as2d(n)
+           dsvs(i,j)  =corp(j,1,n)*as2d(n)
```
shoyokota: Here only corp( : , 1 , : ) is used, so corp( : , 2:3 , : ) is not used at all although it is input. Is this behavior as you expect?

src/gsi/prewgt_reg.f90
```diff
@@ -359,8 +376,8 @@ subroutine prewgt_reg(mype)
                   '(PREWGT_REG:pe=',mype,                                                    &
                   ') re-set the length-scale(hor ver) of vp (VelPotent): ',                  &
                   hscl_vp, vscl_vp,'   n=', n,cvars3d(n)
-              if ( hscl_vp > 0.0_r_kind ) hwll(:,:,n) = hscl_vp
-              if ( vscl_vp > 0.0_r_kind ) vz(  :,:,n) = vscl_vp
+              if ( hscl_vp > 0.0_r_kind ) hwll(:,:,1,n) = hscl_vp
+              if ( vscl_vp > 0.0_r_kind ) vz(  :,:,1,n) = vscl_vp
 
           end if
```
shoyokota, line 381: Why didn't you adjust corz/hwll/vz for all bins?

src/gsi/prewgt_reg.f90
```diff
@@ -378,38 +395,38 @@ subroutine prewgt_reg(mype)
...
           if (n==nrf3_w ) then
               if (mype==0) &
                   write(6,'(1x,A60,I4)')'PREWGT_REG: user-defined namelist to tune BE for w on pe:',mype
-                                              corz(:,:,n) = 3.0_r_kind
-              if (  hscl_qx > 0.0_r_kind ) hwll(:,:,n) = hscl_qx
-              if (  vscl_qx > 0.0_r_kind ) vz(  :,:,n) = vscl_qx
+                                              corz(:,:,1,n) = 3.0_r_kind
+              if (  hscl_qx > 0.0_r_kind ) hwll(:,:,1,n) = hscl_qx
+              if (  vscl_qx > 0.0_r_kind ) vz(  :,:,1,n) = vscl_qx
           end if
```
shoyokota, line 429: Why didn't you adjust corz/hwll/vz for all bins?

src/gsi/prewgt_reg.f90
```diff
@@ -473,13 +508,13 @@ subroutine prewgt_reg(mype)
...
             do j=llmin,llmax
-                call smoothzo(vz(1,j,n),samp,rate,n,j,dsv(1,1,j))
+                call smoothzo(vz(1,j,1,n),samp,rate,n,j,dsv(1,1,1,j))
             end do
 
             do j=1,lat2
                 mid_nsig = int((nsig+1)/2)
                 mid_mlat = int((llmin+llmax)/2)
-                vz_cld(1:nsig) = vz(1:nsig,mid_mlat,n)    !
+                vz_cld(1:nsig) = vz(1:nsig,mid_mlat,1,n)    !
```
shoyokota, line 508: vz( : , : , 2:3 , : ) is not used here. Is this behavior as you expect?

src/gsi/balmod.f90
```diff
@@ -399,6 +421,15 @@ subroutine prebal_reg(cwcoveqqcov)
     real(r_kind),allocatable,dimension(:,:):: wgvi ,bvi
     real(r_kind),allocatable,dimension(:,:,:):: agvi
 
+    integer(i_kind) num_bins2d,ivar
+    real(r_kind),allocatable,dimension(:,:,:,:)  :: evi2
+    real(r_kind),allocatable,dimension(:,:,:,:,:):: evi3
+
+    integer(i_kind)          :: extra_vartotnum,ii,jj,kk,varinfulllist,istatus
+    logical                  :: if_exist
+    character(len=1024)      :: varname
```
shoyokota: Some of these variables (extra_vartotnum, ii, jj, kk, varinfulllist, if_exist, varname) are not used below.

src/gsi/balmod.f90
```diff
@@ -483,9 +528,63 @@ subroutine prebal_reg(cwcoveqqcov)
...
+                         l2=min0(l+1,llmax)
+                         dl2=rllat1(i,j)-float(l)
+                         dl1=one-dl2
+                         if( ges_mask(i,j,k) < -0.5_r_kind   )then
+                            evik2_real(ivar,i,j,k)=dl1*evi2(ivar,l,2,k)+dl2*evi2(ivar,l2,2,k)
+                         else if ( ges_mask(i,j,k) > 0.5_r_kind  )then
+                            evik2_real(ivar,i,j,k)=dl1*evi2(ivar,l,3,k)+dl2*evi2(ivar,l2,3,k)
+                         else
+                            evik2_real(ivar,i,j,k)=dl1*evi2(ivar,l,1,k)+dl2*evi2(ivar,l2,1,k)
+                         end if
```
shoyokota, line 547: Considering that corp( : , 2:3 , : ) is not used in prewgt_reg.f90, evik2_real might not have to be binned. The consistency with evik3_real is also questionable as the following comment.

src/gsi/balmod.f90
```diff
@@ -483,9 +528,63 @@ subroutine prebal_reg(cwcoveqqcov)
...
+                         l2=min0(l+1,llmax)
+                         dl2=rllat1(i,j)-float(l)
+                         dl1=one-dl2
+                         if( ges_mask(i,j,k) < -0.5_r_kind  )then
+                            evik3_real(ivar,i,j,m,k)=dl1*evi3(ivar,l,2,m,k)+dl2*evi3(ivar,l2,2,m,k)
+                         else if ( ges_mask(i,j,k) > 0.5_r_kind  )then
+                            evik3_real(ivar,i,j,m,k)=dl1*evi3(ivar,l,3,m,k)+dl2*evi3(ivar,l2,3,m,k)
+                         else
+                            evik3_real(ivar,i,j,m,k)=dl1*evi3(ivar,l,1,m,k)+dl2*evi3(ivar,l2,1,m,k)
+                         end if
```
shoyokota, line 570: If you decide to bin evik2_real(:,i,j,k) by ges_mask(i,j,k), I think evik3_real(:,i,j,m,k) should be binned both by ges_mask(i,j,k) and by ges_mask(i,j,m). Why did you use only ges_mask(i,j,k)?

src/gsi/control2state.f90
```diff
@@ -484,6 +494,13 @@ subroutine c2sset(xhat,sval)
 do_normal_rh_to_q=lc_rh.and.lc_t .and.ls_prse.and.ls_q
 do_tv_to_tsen    =lc_t .and.ls_q .and.ls_tsen
 do_getuv         =lc_sf.and.lc_vp.and.ls_u.and.ls_v
+do_uv_copy=.false.
```
shoyokota: 'do_uv_copy=.false.' is not needed.

src/gsi/ensctl2state_ad.f90
shoyokota, line 1: This file doesn't have to be modified because lc_u, lc_v, and do_uv_copy are not used here.

src/gsi/m_berror_stats_reg.f90
```diff
@@ -290,6 +326,801 @@ subroutine berror_read_bal_reg(msig,mlat,agvi,bvi,wgvi,mype,unit)
...
+      real(r_kind),dimension(extra_var2num,0:mlat+1,num_bins2d,nsig)     ,intent(  out),optional :: evi2
+      real(r_kind),dimension(extra_var3num,0:mlat+1,num_bins2d,nsig,nsig),intent(out) :: evi3
+      integer(i_kind)                              ,intent(in   ) :: mype  !"my" processor ID
+      integer(i_kind),optional                     ,intent(in   ) :: unit !analternative unit
+
+      ! LOCAL DECLARE
+      integer(i_kind)          :: &
+                                  ii,jj,kk
+      logical                  :: if_exist
+      character(len=1024)      :: varname
```
shoyokota: These variables (ii, jj, kk, if_exist, varname) are not used below.

src/gsi/m_berror_stats_reg.f90
```diff
@@ -290,6 +326,801 @@ subroutine berror_read_bal_reg(msig,mlat,agvi,bvi,wgvi,mype,unit)
...
+  real(r_kind) :: factoz
+  real(r_kind) :: raux
+
+  real(r_kind), parameter :: corz_default=one,hwll_default=100000_r_kind,&
+                             vz_default=one
+  real(r_kind) ,dimension(mlat,1,2) :: cov_dum
+
+  integer(i_kind)          :: ii, jj, kk
+  logical                  :: if_exist
+  character(len=1024)      :: varname
```
shoyokota: These variables (ii, jj, kk, if_exist, varname) are not used below.

shoyokota (commented):
I roughly checked the code. This development to introduce the new static B configuration is very interesting, but I found some questionable parts affecting the quality of the analysis. Could you check them?

src/gsi/prewgt_reg.f90
YueYang-OU: > The print to log file is not needed here.

Removed.

src/gsi/prewgt_reg.f90
YueYang-OU: > This bin_l is not used. I think it can be deleted.

This part includes bugs when we merged the codes. Please see the latest codes.

src/gsi/prewgt_reg.f90
YueYang-OU: > Here only hwll( : , : , 1 , : ) is used, so hwll( : , : , 2:3 , : ) is not used at all although it is input. Is this behavior as you expect?

We use bin=2 for storm areas.

src/gsi/m_berror_stats_reg.f90
YueYang-OU: > For the case of qoption=2, I think varq also should be varq( 1:mlat , 1:nsig , 1:num_bins2d ) and binned by reflectivity.

When computing the static statistics, qoption=1 is mandatory and varq is not used.

src/gsi/m_berror_stats_reg.f90
YueYang-OU: > For the case of cwoption=2, I think varcw also should be varcw( 1:mlat , 1:nsig , 1:num_bins2d ) and binned by reflectivity.

varcw is not used.

src/gsi/balmod.f90
YueYang-OU: > I think these subroutines (det_field, balance_extra, and tbalance_extra) should be generalized to be able to apply for any number of nb_cv and any kinds of 3D and 2D variables.

Thanks for the suggestions. We agree with that. However, iterative computation is required in the balance part. We will think about it and add the capability later.

src/gsi/directDA_radaruse_mod.f90
YueYang-OU: > Why did you change the default l_plt_be_stats?

It takes a while to create a file with the default l_plt_be_stats = .True. and the created file is not always useful to all users.

src/gsi/gsi_rfv3io_mod.f90
YueYang-OU: Agree and will do later.

src/gsi/prewgt_reg.f90
YueYang-OU: > Here only corp( : , 1 , : ) is used, so corp( : , 2:3 , : ) is not used at all although it is input. Is this behavior as you expect?

We didn't include these 2D control variables in the developed static BEC.

src/gsi/prewgt_reg.f90
YueYang-OU: > Why didn't you adjust corz/hwll/vz for all bins?

This part is related to another static BEC option, which was developed by the other group. We only extended the dims of these arrays for consistency.

src/gsi/balmod.f90
YueYang-OU: > If you decide to bin evik2_real(:,i,j,k) by ges_mask(i,j,k), I think evik3_real(:,i,j,m,k) should be binned both by ges_mask(i,j,k) and by ges_mask(i,j,m). Why did you use only ges_mask(i,j,k)?

ges_mask is defined based on the composite reflectivity, so it is fine to use ges_mask(i,j,k).

src/gsi/prewgt_reg.f90
shoyokota: I saw the latest code, but why is bin_l=2 selected in if_cs_staticB?

src/gsi/prewgt_reg.f90
shoyokota: It seems that bin_l=2 is used even except for storm areas. Why?

src/gsi/m_berror_stats_reg.f90
shoyokota: In gsiparm.anl you gave me previously, qoption=2 is set. Which is correct? If qoption=1 is required, qoption=1 should be forced somewhere and this "if" part should be removed.

src/gsi/m_berror_stats_reg.f90
shoyokota: If cwoption=1 is required, cwoption=1 should be forced somewhere and this "if" part should be removed.

src/gsi/bkerror.f90
shoyokota: Why did you add only "dbz"?

src/gsi/directDA_radaruse_mod.f90
shoyokota: I agree with that, but changing the default value often causes unexpected confusion in the system using this GSI. I don't think it should be changed for now.

src/gsi/gsi_rfv3io_mod.f90
shoyokota: It should be enclosed not within 'if( if_consistency_ratio )' but within 'if( if_cs_staticB .and. if_consistency_ratio )' here.

src/gsi/prewgt_reg.f90
shoyokota: While it is not binned by corp, evik2_real is binned by evi2 in balmod.f90. It is not consistent. Please unify it.

src/gsi/prewgt_reg.f90
shoyokota: If so, I think this part should be enclosed within 'if( if_cs_staticB )'.

src/gsi/prewgt_reg.f90
shoyokota: I think this part should be enclosed within 'if( if_cs_staticB )'.

src/gsi/balmod.f90
shoyokota: Does it mean that ges_mask is one of the 2D variable? The GSI-mask code doesn't look so, but if it is exactly the 2D variable, ges_mask should not depend on k and it should be ges_mask( : , : ).

src/gsi/m_berror_stats_reg.f90
```diff
@@ -290,6 +326,817 @@ subroutine berror_read_bal_reg(msig,mlat,agvi,bvi,wgvi,mype,unit)
...
+     if (n==nrf2_pmsl) then
+        do i=1,mlat
+           corp(i,:,n)=corp(i,:,nrf2_ps)
+        end do
+        do i=0,mlat+1
+           hwllp(i,:,n)=hwllp(i,:,nrf2_ps)
+        end do
+     end if
+     if (n==nrf2_howv) then
+        call read_howv_stats(mlat,1,2,cov_dum)
```
shoyokota: It should be changed to read_howv_stats(mlat,1,2,cov_dum,**mype**) in the latest code.

src/gsi/prewgt_reg.f90
wangym1111: This is a temporary option for the recursive filter, where the same length scale is used in the entire domain. Because we use convective-scale static B in the storm-scale assimilation only, it is still safe to use the weak-storm length scale for the entire domain.

src/gsi/prewgt_reg.f90
wangym1111: See the above

src/gsi/m_berror_stats_reg.f90
wangym1111: qoption=1 is correct. We will add the safeguard later.

src/gsi/m_berror_stats_reg.f90
wangym1111: We are not sure if the failure you mentioned is caused by the mismatch of vertical levels between static B and background files. We didn't see sigma_avn(k)=zero and will double-check it.

src/gsi/directDA_radaruse_mod.f90
wangym1111: Reverted it.

src/gsi/prewgt_reg.f90
wangym1111: Changed.

src/gsi/prewgt_reg.f90
wangym1111: These codes are not relevant to the convective-scale static B. We changed this to keep the consistency of array dimensions throughout the code.

src/gsi/balmod.f90
wangym1111: An NCL code is used in the shell script to process the mask file further to ensure the use of composite reflectivity. We are developing a Python script to replace the GSI-mask code. This Python script may avoid confusion. Given all levels have the same values in ges_mask, the current code still works.

src/gsi/m_berror_stats_reg.f90
wangym1111: Corrected.

src/gsi/prewgt_reg.f90
shoyokota: If so, bin_l=1 also should be safe. Why is it required to change it from 1 to 2?

src/gsi/m_berror_stats_reg.f90
shoyokota: I checked again with the xxd command and confirmed sigma_svn(k)=zero for k>55 in the convective-scale static B that I received in spite of msig=65.

src/gsi/prewgt_reg.f90
shoyokota: Sorry, let me modify my comment as follows.

> If so, I think this part should be enclosed within 'if( **.not.** if_cs_staticB )'.

src/gsi/prewgt_reg.f90
shoyokota: Sorry, let me modify my comment as follows.

> I think this part should be enclosed within 'if( **.not.** if_cs_staticB )'.

src/gsi/balmod.f90
shoyokota: It is correct, but the current GSI code doesn't work properly unless all levels have the same values in ges_mask. I think it causes confusion even if the python script is used instead of the NCL code. If it is impossible to make ges_mask the 2D variable, I suggest binning as follows in GSI. Similar modification might be required also for ges_mask1.
```
if( maxval(ges_mask(i,j,:)) > 90.0_r_kind )then
   (use 3rd bin)
else if ( minval(ges_mask(i,j,:)) < -90.0_r_kind )then
   (use 2nd bin)
else
   (use 1st bin)
endif
```

src/gsi/cplr_get_fv3_regional_ensperts.f90
```diff
@@ -452,7 +452,7 @@ subroutine get_fv3_regional_ensperts_run(this,en_perts,nelen,ps_bar)
 
               select case (trim(cvars3d(ic3)))
  
-                 case('sf','SF')
+                 case('sf','SF','u','U')
```
shoyokota, line 455: Modification to use u/v in calculation of ensemble B is required not only in this file but also in ensctl2state.f90, ensctl2state_ad.f90, ensctl2model.f90, ensctl2model_ad.f90 and write_fv3_spread.f90.

shoyokota:
@YueYang-OU @wangym1111 I heard that the GSI handling reviewer team is thinking about changing this PR to draft mode because this PR is beyond the 6 week limit. This PR is very helpful for me to run the experiment with this convective-scale static B, but I think it is still not on the stage to merge to the develop branch, and not planned to be used in RRFSv1 at this moment. Could you let us know the current status of your modification? Do you think this PR should be merged as soon as possible?

RussTreadon-NOAA:
Close this PR with the possibility of reopening at a later date.  The originating issue, #576, will remain open.

PR #594, opened 2023-07-26
Optimize the reading of ensembles and setup for global multiscale runs

### LINKED ISSUE #585: GSI optimization with focus on ensemble input and multiscale setup.
The GSI spends a lot of wall time on reading in the ensembles and setting up the multiscale ensemble setup when these options are used.  In these changes, the code is optimized to speed up these two steps of the analysis.

The changes made to the analysis are:
1. Keep ensembles single precision throughout the cplr step for the ensembles.
Note this does not result in any change in the result since the variables were changed to double precision right before the update_halos_ routine (step removed).  A few things such as conversion from sensible temperature to virtual temperature were moved to later so that the results were not impacted by the change in precision.
2. In the creation of the multiscale ensembles, the conversion from grid to spectral was done only once for all scales.  Before it was done for each scale thus eliminating nsclgrp-1 sub2grids and conversion to spectrals.
3. In the creation of the multiscale ensembles, the sub2grid and grid2sub was converted from double to single precision reducing the amount of data passed between processors.  Note there is no loss of information because the data is in single precision immediately before and after the sub2grid and grid2sub routines.
4. The calculation of the weights for the multiscale ensembles and spectral filter was optimized.
5. A specialized sub2grid and grid2sub was created to eliminate some unnecessary data movement.
6. The update_halos routine was removed, the read ensemble routines modified and the input parameters modified for the genex routines to ensure that the boundary points of the subdomains were specified when genex called.  By eliminating the update_halos routine a lot of all-to-all communications were removed.  Also significant simplification of code results. This impacts both multiscale runs and non-multiscale runs.
7. Some threading and restructuring of loops was done in hybrid_ensemble_isotropic.F90, especially for the background errors for the ensembles.
8. The calculation of the RH in get_gefs_ensperts_dualres.f90 was simplified and optimized.

All of the above changes can be made with bit identical results.  However, two minor improvements to the code were made that do result in a minor change to the results.
A. The moisture is not checked for negative values before the virtual temperature is calculated in the original code.
B. To get sensible temperature, the old code converts it to virtual temperature (with A) and then converts back to sensible temperature.  Using the original sensible temperature results in some minor round-off differences.

The above 2 changes can easily be removed and results in bit-wise identical results.

Note that 2 minor changes to the code were also included which allow the code to pass in debug mode.
a. In read_iasi, the code was changed when the channel was zero to not have it write to that location.
b. In stprad, the abi2km_bc array is only initialized when abi2km and regional.  It could write outside the array when the number of channels was < 4.

When these changes were incorporated the wall times for a multiscale ensemble test I am running went from about 4000-4200s to 2800 - 3100s.  Of course these times are variable, but the new run times seemed to be more consistent from run-to-run.  The multiscale test I am running is not necessarily realistic.  I am using full resolution ensembles and only 2 scales for the multiscale test.  

Test multiscale GSI

            Run time old    Run time new  Max Mem old Max Mem new
1              4384s              2980s         90605476MB  99611360MB
2              4058s              3115s         98866156MB  90699556MB
3              4127s              3159s         98338100MB  91108816MB
4              4179s              2743s         97168444MB  100365004MB

With lower resolution the impact should be less.  With more scales the impact should be larger. I am working with Cathy to test in a more realistic situation.  Some results are posted by Cathy below.  Thanks Cathy!

The changes should also impact the run time of the global GSI as it is being run in operations.  I do not believe it will impact any of the other current operational runs.

For the global GSI, the run times for the global are:

            Run time old    Run time new  Max Mem old Max Mem new
1              1829s              1829s         75373748MB  73627272MB
2              1804s              1663s         70815928MB  69961644MB
3              1781s              1767s         75953952MB  76127660MB
4              1849s              1644s         69545452MB  68708012MB

In general the test case shows faster times, but there is still significant run-to-run variability.  

Changes are in my optimize3 branch.  All testing was done on Hera.  All runs are done with --mem=0

jderber-NOAA:
This update improves the efficiency of the GSI, especially for multiscale runs.  Details can be found in issue#585

The runs produce identical results except when ensembles are used.  Identical results can be produced with ensembles as well with changes to 3 lines of code.  These lines zero out negative moistures before creating virtual temperatures and use the original sensible temperatures rather than ones created from virtual temperatures (which were created from the original sensible temperatures).

All regression tests passed except due to the above reason.  When those 3 lines were changed back, all regression tests passed.  Changes due to the above 3 lines were very minor.

All testing was performed by myself on Hera.  

See Issue to see examples of speed-ups of the code that resulted from this change.
  
Fixes #585 

**Checklist**

- [x ] My code follows the style guidelines of this project
- [x ] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [x ] New and existing tests pass with my changes
- [x] Any dependent changes have been merged and published

**DUE DATE for this PR is 9/6/2023.** If this PR is not merged into develop by this date, the PR will be closed and returned to the developer.


src/gsi/read_iasi.f90
```diff
@@ -827,7 +827,11 @@ subroutine read_iasi(mype,val_iasi,ithin,isfcalc,rmesh,jsatid,gstime,&
 !          Put satinfo defined channel temperatures into data array
            do l=1,satinfo_nchan
               i = bufr_index(l)
-              data_all(l+nreal,itx) = temperature(i)   ! brightness temerature
+              if(i /= 0)then
+                 data_all(l+nreal,itx) = temperature(i)   ! brightness temerature
+              else
```
TingLei-NOAA, line 834: Should we put a warning printout here to indicate i=0  and tbmin is used?

src/gsi/setuprad.f90
```diff
@@ -773,6 +773,12 @@ subroutine setuprad(obsLL,odiagLL,lunin,mype,aivals,stats,nchanl,nreal,nobs,&
      call stop2(275)
   end if
 
+  if (abi2km .and. regional) then
+     abi2km_bc = zero
+     abi2km_bc(2) = 233.5_r_kind
+     abi2km_bc(3) = 241.7_r_kind
+     abi2km_bc(4) = 250.5_r_kind
+  end if
```
TingLei-NOAA, line 784: Is this block not needed to global runs?

jderber-NOAA: The original code will not use these constants unless regional and abi2km are both true.  I do not want to change the operation of the code without testing.  Just moved the setting of these constants outside of loop and only set when they can be used.

src/gsi/read_iasi.f90
jderber-NOAA: I think I convinced myself that the data would not be used if i=0.  So I don't think any print out is necessary.

jderber-NOAA: Yes.  In the qc routines if the data is equal to tbmin, the obs errors are set so that it will not be used.

jderber-NOAA:
Updated branch with changes to trunk.

src/gsi/cplr_gfs_ensmod.f90
```diff
@@ -443,19 +435,25 @@ subroutine move2bundle_(grd3d,sloc,atm_bundle,m_cvars2d,m_cvars3d,iret)
...
           write(6,'(A)') trim(myname_) // ': but some have not been found. Aborting ... '
           write(6,'(A)') trim(myname_) // ': WARNING!'
-          write(6,'(3A,I5)') trim(myname_) // ': Trouble reading ensemble file : ', trim(filename), ', IRET = ', iret
        endif
        return
     endif
 
+    do m=1,nc2d
+!      convert ps from Pa to cb
+       if(trim(cvars2d(m))=='ps') ps=en_loc3(:,:,m_cvars2d(m))
```
CatherineThomas-NOAA, line 444: Is the removal of the r0_001 factor taken into account somewhere else?

ush/build.sh
```diff
@@ -6,6 +6,7 @@ set -eux
 readonly DIR_ROOT=$(cd "$(dirname "$(readlink -f -n "${BASH_SOURCE[0]}" )" )/.." && pwd -P)
 
 # User Options
+# BUILD_TYPE=${BUILD_TYPE:-"Debug"}
```
CatherineThomas-NOAA: Was this unintended?

src/gsi/cplr_gfs_ensmod.f90
CatherineThomas-NOAA: Nevermind, I found it in get_gefs_ensperts_dualres

jderber-NOAA: Right!  It was moved so that the data could stay in single precision until it got to get_gefs_ensperts_dualres.  Thanks!

ush/build.sh
jderber-NOAA: The line is commented out, but should be removed.  I was testing the code in debug mode as well!

CatherineThomas-NOAA (dismissed):
Looks good to me! (minus the one comment for build.sh)  My latest timings are on Issue #585.  Thanks @jderber-NOAA!

jderber-NOAA:
Updated to head of trunk and remove commented out line from build.sh.

jderber-NOAA:
Regression tests were rerun with this updated version of the code.  All regression tests passed except 4denvar.  The last update to the trunk appears to have introduced a very small change into the 4denvar.  This difference is certainly at the scale of round-off.   No difference in the initial penalty.  First 3 iterations of the control and update are given below.

< cost,grad,step,b,step? =   1   0  6.592319865747931181E+05  1.700824233239468640E+03  1.057506532852620307E+00  0.000000000000000000E+00  good
< cost,grad,step,b,step? =   1   1  6.555948920937654329E+05  2.114103169411071576E+03  1.927472355822770655E+00  1.302893266057407962E+00  good
< cost,grad,step,b,step? =   1   2  6.469553380174754420E+05  1.349404725181090953E+03  2.810796397253191525E+00  1.212777155871386014E+00  good

> cost,grad,step,b,step? =   1   0  6.592319865747931181E+05  1.700824233239468640E+03  1.057506532852620751E+00  0.000000000000000000E+00  good
> cost,grad,step,b,step? =   1   1  6.555948920937654329E+05  2.114103169411072486E+03  1.927472355822761774E+00  1.302893266057409294E+00  good
> cost,grad,step,b,step? =   1   2  6.469553380174754420E+05  1.349404725181082085E+03  2.810796397253162660E+00  1.212777155871367807E+00  good

Trying to find reason for small difference.

RussTreadon-NOAA:
**WCOSS2 ctests**
Install `jderber-NOAA:optimize3` at [`3e918e1`](https://github.com/jderber-NOAA/GSI/commit/3e918e1967f9ee1a0c7aa3c63efbadb12ccaba04) on Dogwood.  Run ctests with following results
```
russ.treadon@dlogin08:/lfs/h2/emc/da/noscrub/russ.treadon/git/gsi/pr594/build> ctest -j 9
Test project /lfs/h2/emc/da/noscrub/russ.treadon/git/gsi/pr594/build
    Start 1: global_3dvar
    Start 2: global_4dvar
    Start 3: global_4denvar
    Start 4: hwrf_nmm_d2
    Start 5: hwrf_nmm_d3
    Start 6: rtma
    Start 7: rrfs_3denvar_glbens
    Start 8: netcdf_fv3_regional
    Start 9: global_enkf
1/9 Test #8: netcdf_fv3_regional ..............***Failed  483.81 sec
2/9 Test #7: rrfs_3denvar_glbens ..............   Passed  605.92 sec
3/9 Test #9: global_enkf ......................   Passed  678.08 sec
4/9 Test #5: hwrf_nmm_d3 ......................   Passed  797.90 sec
5/9 Test #4: hwrf_nmm_d2 ......................   Passed  1026.16 sec
6/9 Test #6: rtma .............................   Passed  1272.35 sec
7/9 Test #3: global_4denvar ...................***Failed  1503.70 sec
8/9 Test #2: global_4dvar .....................   Passed  1744.93 sec
9/9 Test #1: global_3dvar .....................   Passed  1983.04 sec

78% tests passed, 2 tests failed out of 9

Total Test time (real) = 1983.04 sec

The following tests FAILED:
          3 - global_4denvar (Failed)
          8 - netcdf_fv3_regional (Failed)
Errors while running CTest
```
The `netcdf_fv3_regional` failure is due to 
```
The memory for netcdf_fv3_regional_loproc_updat is 352488 KBs.  This has exceeded maximum allowable memory of 238774 KBs, resulting in Failure memthresh of the regression test.
```
A check of the task 0 maximum resident set sizes for the updat (`jderber-NOAA:optimize3`) and contrl (`develop`) confirms that the `loproc_updat` uses more memory than `loproc_contrl`
```
netcdf_fv3_regional_hiproc_contrl/stdout:The maximum resident set size (KB)                   = 364344
netcdf_fv3_regional_hiproc_updat/stdout:The maximum resident set size (KB)                   = 364280
netcdf_fv3_regional_loproc_contrl/stdout:The maximum resident set size (KB)                   = 217068
netcdf_fv3_regional_loproc_updat/stdout:The maximum resident set size (KB)                   = 352488
```
The `loproc_updat` maximum resident set size is more consistent with the `loproc_contrl` for other ctests.  It's not clear why the difference is larger for `netcdf_fv3_regional`.   This failure, however, is not viewed as a fatal fail.

The `global_4denvar` failure is due to 
```
The results (penalty) between the two runs are nonreproducible,
thus the regression test has Failed on cost for global_4denvar_loproc_updat and global_4denvar_loproc_contrl analyses.

The case has Failed the scalability test.
The slope for the update (54.378945 seconds per node) is less than that for the control (59.441120 seconds per node).
```
A check of the wall times shows that the `updat` code runs faster than the `contrl`
```
global_4denvar_hiproc_contrl/stdout:The total amount of wall time                        = 278.396694
global_4denvar_hiproc_updat/stdout:The total amount of wall time                        = 262.273809
global_4denvar_loproc_contrl/stdout:The total amount of wall time                        = 337.837814
global_4denvar_loproc_updat/stdout:The total amount of wall time                        = 305.776965
```
This is consistent with the optimization purpose of this PR.  This is not a fatal fail.

The non-reproducible results between the `updat` and `contrl` is more puzzling.  The initial total penalty and gradient are identical between the two codes.   Differences show up in the step size for the second iteration of the first outer loop.  15 of the 19 printed digits are identical.  Differences in the last four digits are at the level of `real(8)` numerical roundoff

`updat`
```
Initial cost function =  6.592319865747931181E+05
Initial gradient norm =  1.700824233239470232E+03
cost,grad,step,b,step? =   1   0  6.592319865747931181E+05  1.700824233239470232E+03  1.057506532852622527E+00  0.000000000000000000E+00  good
cost,grad,step,b,step? =   1   1  6.555948920937654329E+05  2.114103169411077488E+03  1.927472355822775318E+00  1.302893266057414179E+00  good
```
`contrl`
```
Initial cost function =  6.592319865747931181E+05
Initial gradient norm =  1.700824233239470232E+03
cost,grad,step,b,step? =   1   0  6.592319865747931181E+05  1.700824233239470232E+03  1.057506532852622527E+00  0.000000000000000000E+00  good
cost,grad,step,b,step? =   1   1  6.555948920937654329E+05  2.114103169411077488E+03  1.927472355822779093E+00  1.302893266057414179E+00  good
```
John found similar behavior in his tests.

jderber-NOAA:
orking on the nonreproducible issue.  I have not found anything that I changed that should change the round-off or anything.  Hope to find something soon. Surprised at the large increase in memory.  Will look at it after finding the nonreproducible issue.

RussTreadon-NOAA:
@jderber-NOAA , when you have time would you please update `jderber-NOAA:optimize3` with the current head of the authoritative GSI `develop`?  We may need to do this a few more times before this PR is merged into `develop`.

jderber-NOAA:
Reason for reproducibility issue found.  It was documented earlier in this development.  With the change of 3 lines in get_gefs_ensperts_dualres.f90 (around line 190) all regression tests passed.

While looking for the reproducibility problem a few changes were made.  

1. A major error in control_vectors.f90 was found.  The results from partsum were not saved from one value of nsubwin to the next.  By saving the values of partsum, this problem is eliminated.
2. A very minor change was made in general_spectral_transforms.f90.  The lines were already in place but commented out.  real(grd%nlon,r_kind) was used rather than float(grd%nlon).
3. In hybrid_ensemble_isotropic.F90, subroutine bkerror_a_en, the indices of the alphacvarsclgrpmat array were reversed.  This is a symmetric matrix so is not a real issue, but should be made right.
4. In read_prepbufr.f90, subroutine read_prepbufr, the initialization of uob,vob and oelev used 0.0.  This was changed to using constant zero.

Regression tests were rerun.  All passed except rrfs_3denvar_glbens.  Not sure why this did not pass.  Results were the same.
rrfs_3denvar_glbens_hiproc_contrl/stdout:The total amount of wall time                        = 93.271814
rrfs_3denvar_glbens_hiproc_updat/stdout:The total amount of wall time                        = 81.411227
rrfs_3denvar_glbens_loproc_contrl/stdout:The total amount of wall time                        = 135.722101
rrfs_3denvar_glbens_loproc_updat/stdout:The total amount of wall time                        = 111.039343

Run times were faster.
rrfs_3denvar_glbens_hiproc_contrl/stdout:The maximum resident set size (KB)                   = 1136508
rrfs_3denvar_glbens_hiproc_updat/stdout:The maximum resident set size (KB)                   = 1136728
rrfs_3denvar_glbens_loproc_contrl/stdout:The maximum resident set size (KB)                   = 1793072
rrfs_3denvar_glbens_loproc_updat/stdout:The maximum resident set size (KB)                   = 1793692

Slightly more memory.  This must have been the reason for the failure.  But results are reasonable.

Updating to head of trunk.

RussTreadon-NOAA:
**WCOSS2 ctests**
Install fresh clone of `jderber-NOAA:optimize3` at [0b6bde9](https://github.com/jderber-NOAA/GSI/commit/0b6bde994b2ea899fc3ebc92c8c07b85f7897703) on Cactus.  Run ctests with following results
```
russ.treadon@clogin04:/lfs/h2/emc/da/noscrub/russ.treadon/git/gsi/pr594/build> ctest -j 9
Test project /lfs/h2/emc/da/noscrub/russ.treadon/git/gsi/pr594/build
    Start 1: global_3dvar
    Start 2: global_4dvar
    Start 3: global_4denvar
    Start 4: hwrf_nmm_d2
    Start 5: hwrf_nmm_d3
    Start 6: rtma
    Start 7: rrfs_3denvar_glbens
    Start 8: netcdf_fv3_regional
    Start 9: global_enkf
1/9 Test #8: netcdf_fv3_regional ..............***Failed  483.92 sec
2/9 Test #5: hwrf_nmm_d3 ......................   Passed  493.34 sec
3/9 Test #7: rrfs_3denvar_glbens ..............   Passed  605.97 sec
4/9 Test #4: hwrf_nmm_d2 ......................   Passed  607.00 sec
5/9 Test #9: global_enkf ......................   Passed  610.84 sec
6/9 Test #6: rtma .............................   Passed  1210.05 sec
7/9 Test #3: global_4denvar ...................   Passed  1442.23 sec
8/9 Test #1: global_3dvar .....................   Passed  1502.22 sec
9/9 Test #2: global_4dvar .....................   Passed  1563.07 sec

89% tests passed, 1 tests failed out of 9

Total Test time (real) = 1563.08 sec

The following tests FAILED:
          8 - netcdf_fv3_regional (Failed)
```
The `netcdf_fv3_regional` failure is due to the timing scalability check.  
```
The case has Failed the scalability test.
The slope for the update (.601893 seconds per node) is less than that for the control (1.270841 seconds per node).
```
Examination of the _updat_ and _contrl_ wall times does not show any anomalous behavior
```
russ.treadon@clogin04:/lfs/h2/emc/ptmp/russ.treadon/pr594/tmpreg_netcdf_fv3_regional> grep wall */stdout
netcdf_fv3_regional_hiproc_contrl/stdout:The total amount of wall time                        = 63.280665
netcdf_fv3_regional_hiproc_updat/stdout:The total amount of wall time                        = 63.970015
netcdf_fv3_regional_loproc_contrl/stdout:The total amount of wall time                        = 64.551506
netcdf_fv3_regional_loproc_updat/stdout:The total amount of wall time                        = 64.451530
```

This is not a fatal fail.

jderber-NOAA:
After updating to the head of the trunk (both the control and updat), all regression tests passed on Hera.

Test project /scratch1/NCEPDEV/da/John.Derber/converge4/GSI/build
    Start 1: global_3dvar
    Start 2: global_4dvar
    Start 3: global_4denvar
    Start 4: hwrf_nmm_d2
    Start 5: hwrf_nmm_d3
    Start 6: rtma
    Start 7: rrfs_3denvar_glbens
    Start 8: netcdf_fv3_regional
    Start 9: global_enkf
1/9 Test #5: hwrf_nmm_d3 ......................   Passed  560.00 sec
2/9 Test #9: global_enkf ......................   Passed  560.73 sec
3/9 Test #8: netcdf_fv3_regional ..............   Passed  606.70 sec
4/9 Test #7: rrfs_3denvar_glbens ..............   Passed  610.09 sec
5/9 Test #4: hwrf_nmm_d2 ......................   Passed  610.56 sec
6/9 Test #6: rtma .............................   Passed  1456.16 sec
7/9 Test #3: global_4denvar ...................   Passed  1636.21 sec
8/9 Test #2: global_4dvar .....................   Passed  1811.06 sec
9/9 Test #1: global_3dvar .....................   Passed  1813.70 sec

100% tests passed, 0 tests failed out of 9

Total Test time (real) = 1813.71 sec


I will put back in the code the 3 lines that do give a minor difference.  These lines result in fewer conversions between variables and should produce slightly more consistent results.

RussTreadon-NOAA (approved):
Approve pending peer reviews

TingLei-NOAA (approved):
Finished another a test using 3km conus domain RRFS case.  This PR gives the identical results (final cost and gradients)  compared with the EMC GSI trunk . 
Thanks for this continual improvement over GSI

src/gsi/hybrid_ensemble_isotropic.F90
```diff
@@ -3624,34 +3634,29 @@ subroutine bkerror_a_en(grady)
...
-     else
-        allocate(z(naensgrp,nval_lenz_en))
+     end do
+  else
+     allocate(z(nval_lenz_en,naensgrp))
+     allocate(z2(nval_lenz_en))
+     do ii=1,nsubwin
         do ig=1,naensgrp
-           call ckgcov_a_en_new_factorization_ad(ig,z(ig,:),grady%aens(ii,ig,1:n_ens))
+           call ckgcov_a_en_new_factorization_ad(ig,z(1,ig),grady%aens(ii,ig,1:n_ens))
```
CatherineThomas-NOAA, line 3646: @jderber-NOAA Is using z(1,ig) here instead of z(:,ig) intentional?

TingLei-NOAA:
A note: in my RRFS run,  it is found there are little differences between the GSI of this PR built with or not with debug mode. 
For example ( after 1hr30min run), the final cost and grad and so on are , for gsi in debug mode:
```
 cost,grad,step,b,step? =   2  31  5.108161372585962818E+04  4.728171079078897776E+01  2.902662923838958076E+00  1.425154427382457012E+00  good
 ```
It is , for GSI built with "realease" mode:
```
 cost,grad,step,b,step? =   2  31  5.108161597988699941E+04  4.728289846049271006E+01  2.902840431607158767E+00  1.425206942448039582E+00  good
 ```
 EMC GSI trunk shows the same behavior. Namely, in debug mode, this PR and EMC GSI trunk show the identical results for build type : Release or debug and the above tiny differences exist between Release mode and debug mode for both branches.   
 I didn't notice this behavior before.  If it is an issue to be further investigated , it is not specific to this PR.

src/gsi/hybrid_ensemble_isotropic.F90
jderber-NOAA: Good question!!!

Yes, it actually was.  What I will say here is from my knowledge from a few (maybe more) years ago.  It may have changed since then.

Both should result in the same answer.  However when the ":" is used the compiler will have it create a copy of the data before passing it to the subroutine.  When "1" is used the compiler will just pass the starting address to the subroutine.  The second is faster if the data is already sequential.  If the ":" is, for example, in the second dimension, it may be faster because it will bring the data into a new array and will make the data sequential in memory in the subroutine.

CatherineThomas-NOAA: Thanks @jderber-NOAA!

CatherineThomas-NOAA (approved):
Looks great.  Thanks @jderber-NOAA!

@TingLei-NOAA Thanks for looking into that reproducibility issue.  If this is also an issue in the develop branch, I don't see a need to hold up this PR.

TingLei-NOAA:
@CatherineThomas-NOAA  Agree!

jderber-NOAA:
After latest update to head of the trunk, regression test run.  Results as expected.  The update did not impact changes.

RussTreadon-NOAA:
Given the following
- `jderber-NOAA:optimize3` at [982425d](https://github.com/jderber-NOAA/GSI/commit/982425dec79cde79e7b67f28cb58bf35df2bf605) is up to date with authoritative `develop`
- approvals from two peer reviews
- ctests from current head of `jderber-NOAA:optimize3` pass
- GSI Handling Review notified.  OK received

proceed to merge `jderber-NOAA:optimize3` into authoritative `develop`

PR #568, opened 2023-05-04
Feature/enkf q2

TingLei-daprediction:
Resubmit PR for cleaning the unneeded item in the previous PR 
In correspondence to the EMC GSI Issue#566, this PR contains a quick adding of the clipping of negative values of sphum (q) in the analysis of FV3-LAM EnKF.
This part of codes are not tested in the current GSI regression tests, which, hence, are not run.
The current codes are verified using local FV3-LAM case. It is found the differences from this changes exist for sphum ( maximum values about 0.003 (units) and 0.3 K for T (sensible T). The latter is because the sphum would be used when the analysis TV is converted to T. All differences are on spontaneous points and values are reasonable as expected.
Hence, the code is regarded as verified.

Fixes #566

**DUE DATE for this PR is 6/15/2023.** If this PR is not merged into `develop` by this date, the PR will be closed and returned to the developer.

PR #602, opened 2023-07-28
Bugfix/pseudo rh

### LINKED ISSUE #561: EnKF use_qsatensmean and pseudo_rh reverts humidity control variable to q
In the EnKF, the global namelist includes: 

pseudo_rh = true (use RH = q/qsat as humidity control variable) 
use_qsatensmean = true (when calculating q/qsat for control variable, use ensemble mean qsat for all ensemble members) 
write_ensmean = true (write out the ensemble mean, instead of each member).

@jswhit recalls that use_qsatensmean was introduced to prevent issues with very small qsat in some ensemble members. 

If use_qsatensmean is true, in controlvec.f90 read_control (then write_control) the q for each ensemble member are divided (then multiplied) by the same number at each location before (then after) the state update. This multipilcation/division of each ensemble member by the same number has no not effect. Hence, use_qsatensmean = true reduces to being the same as use pseudo_rh = false (using q as the control variable). 

In tests with the LETKF, I got identical analysis from assimilation sondes using [pseudo_rh = false] and using [use_qsatensmean = true and pseudo_rh = true]. However, using [use_qsatensmean = false and pseudo_rh = true] produces a different analysis (that is consistent with using RH as the control variable). 

To address this, I propose: 

1) Eventually, removing the use_qsatensmean option. 

@jswhit Are there other EnKF-like solvers that might behave differently? The reason the qsatmean scaling cancels out is because all the info used in the Hx, HB, O-F was calculated in the observer. It's only the input ens-background and ens-analysis that is scaled by qsat in the enkf code. 

2) To test removing the use_qsatensmean option, add the ability to write out the ensemble mean states when using use_qsatensmean=false. At the moment, writing out the ensemble mean (increment or analysis) requires use_qsatensmean=true. To fix this, grdin in write control will need to be scaled by the ensemble member qsat prior to the mpi call used to calculate the mean.

ClaraDraper-NOAA:
**Description**

The EnKF uses the pseudo_RH=True option, which switches the humidity analysis variable to q/qsat (pseudo_RH=False would use q) 
The default (and in operations) settings also includes use use_qsatensmean=True, which uses q/ensmean(qsat) for each ensemble member.

If use_qsatensmean is true, in controlvec.f90 read_control (then write_control) the q for each ensemble member are divided (then multiplied) by the same number (ensemble mean qsat) at each location before (then after) the state update. This multipilcation/division of each ensemble member by the same number has no net effect, and use_qsatensmean = true reduces to being the same as use pseudo_rh = false (using q as the control variable). Experiments show identical output from both options. See example in slides: 

https://docs.google.com/presentation/d/1mLenEtuoN_9PI0--BEyIHN-E3rUW2WwcsHgzgqi2r0M/edit?usp=sharing

This PR then removes the use_qsatensmean option (including from the namelists). This required adding code to write out the ensemble mean for the pseudo_RH option without relying on the ensemble mean qsat. 

In the PR  I have the changed the pseudo_RH option to False in the regression test, which amounts to no science change. However, True would be better option and is what we thought we were doing up to now (more consistent with what is done in the Var solver, less heterogeneity in ensemble covariances with temperature). I suggest that we switch this going forward.

I ran both cases (but also decreased the obs error) in an EnKF experiment. There was no significant impact on skill. See slides linked above.

Note: @jswhit introduced the use_qsatensmean option to avoid problems with very small qsat at some locations. Could this become a problem again if we revert back to pseudo_RH=True.

Fixes #561  

**Type of change**

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

**How Has This Been Tested?**

See experiments in slides above. 
All regerssion tests have passed on hera. 

  
**Checklist**

- [x ] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] New and existing tests pass with my changes
- [x ] Any dependent changes have been merged and published

**DUE DATE for this PR is 9/8/2023.** If this PR is not merged into develop by this date, the PR will be closed and returned to the developer.

### DIFF
--- a/src/enkf/controlvec.f90
+++ b/src/enkf/controlvec.f90
@@ -64,7 +64,6 @@ private
 public :: read_control, write_control, controlvec_cleanup, init_controlvec
 real(r_single), public, allocatable, dimension(:,:,:,:) :: grdin
 real(r_double), public, allocatable, dimension(:,:,:,:) :: qsat
-real(r_double), public, allocatable, dimension(:,:,:) :: qsatmean
 
 integer(i_kind), public :: nc2d, nc3d, ncdim
 character(len=max_varname_length), allocatable, dimension(:), public :: cvars3d
@@ -192,7 +191,6 @@ subroutine read_control()
 ! read ensemble members on IO tasks
 implicit none
 real(r_double)  :: t1,t2
-real(r_double), allocatable, dimension(:) :: qsat_tmp
 integer(i_kind) :: nb,nlev,ne
 integer(i_kind) :: q_ind
 integer(i_kind) :: ierr
@@ -373,7 +371,6 @@ if (allocated(clevels)) deallocate(clevels)
 if (allocated(index_pres)) deallocate(index_pres)
 if (allocated(grdin)) deallocate(grdin)
 if (allocated(qsat)) deallocate(qsat)
-if (allocated(qsatmean)) deallocate(qsatmean)
 call gridinfo_cleanup()
 end subroutine controlvec_cleanup
 
--- a/src/enkf/controlvec.f90
+++ b/src/enkf/controlvec.f90
@@ -160,7 +160,7 @@ endif
 do i = 1, nc2d
   if (getindex(vars2d_supported, cvars2d(i))<0) then
     if (nproc .eq. 0) then
-      print *,'Error: 2D variable ', cvars2d(i), ' is not supported in current version.'
+      print *,'Error: control 2D variable ', cvars2d(i), ' is not supported in current version.'
       print *,'Supported variables: ', vars2d_supported
     endif
     call stop2(502)
@@ -169,7 +169,7 @@ enddo
 do i = 1, nc3d
   if (getindex(vars3d_supported, cvars3d(i))<0) then
     if (nproc .eq. 0) then 
-       print *,'Error: 3D variable ', cvars3d(i), ' is not supported in current version.'
+       print *,'Error: control 3D variable ', cvars3d(i), ' is not supported in current version.'
        print *,'Supported variables: ', vars3d_supported
     endif
     call stop2(502)
@@ -229,23 +229,23 @@ if (nproc <= ntasks_io-1) then
    end if
    !print *,'min/max qsat',nanal,'=',minval(qsat),maxval(qsat)
    q_ind = getindex(cvars3d, 'q')
-   if (use_qsatensmean .and. q_ind>0 ) then
-       allocate(qsatmean(npts,nlevs,nbackgrounds))
-       allocate(qsat_tmp(npts))
-       ! compute ensemble mean qsat
-       qsatmean = 0_r_double
-       do ne=1,nanals_per_iotask
-          do nb=1,nbackgrounds
-          do nlev=1,nlevs
-             call mpi_allreduce(qsat(:,nlev,nb,ne),qsat_tmp,npts,mpi_real8,mpi_sum,mpi_comm_io,ierr)
-             qsatmean(:,nlev,nb) = qsatmean(:,nlev,nb) + qsat_tmp
-          enddo
-          enddo
-       enddo
-       deallocate(qsat_tmp)
-       qsatmean = qsatmean/real(nanals)
-       !print *,'min/max qsat ensmean',nanal,'=',minval(qsat),maxval(qsat)
-   endif
+   !if (use_qsatensmean .and. q_ind>0 ) then
+   !    allocate(qsatmean(npts,nlevs,nbackgrounds))
+   !    allocate(qsat_tmp(npts))
+   !    ! compute ensemble mean qsat
+   !    qsatmean = 0_r_double
+   !    do ne=1,nanals_per_iotask
+   !       do nb=1,nbackgrounds
+   !       do nlev=1,nlevs
+   !          call mpi_allreduce(qsat(:,nlev,nb,ne),qsat_tmp,npts,mpi_real8,mpi_sum,mpi_comm_io,ierr)
+   !          qsatmean(:,nlev,nb) = qsatmean(:,nlev,nb) + qsat_tmp
+   !       enddo
+   !       enddo
+   !    enddo
+   !    deallocate(qsat_tmp)
+   !    qsatmean = qsatmean/real(nanals)
+   !    !print *,'min/max qsat ensmean',nanal,'=',minval(qsat),maxval(qsat)
+   !endif
    if (nproc == 0) then
      t2 = mpi_wtime()
      print *,'time in readgridata on root',t2-t1,'secs'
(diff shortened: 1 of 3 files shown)


CoryMartin-NOAA (commented):
Code changes look fine, but I will need to test in the global-workflow since the workflow scripts uses use_qsatensmean=.true. in a script that will need to be removed.

jswhit:
I do recall there being crashes when pseudo_RH was True (before use_qsatmean was added).   I wonder if we could/should multiply by the qsat computed from the updated p,T after the EnKF update when pseudo_RH=T?  In that case use_qsatmean=T would not be the same as pseudo_RH=F.

ClaraDraper-NOAA:
@jswhit . I've been thinking about your suggestion, since I do plan to introduce the use of qsat_anal when we scale the RH updated (from pseudo_RH=True) back to q. What your suggestion would effectively do is perform the update on q, then multip the update by qsat_back / qsat_anal. By "perform the update on q", I mean the ensemble correlations would effectively be with q not RH (since every ensemble member is divided by the same qsat). For the assimilation of screen level obs the correlations between q and T are very heterogenous (to the extend that they the sign varies), so using q will likely not give good q and T analyses (for other observations too - but the problem is more acute near the surface). We do need to address the crashing though - I think you said this was happening when q (or spread in q) was too small? Can we do something else to protect against that?

Depending on your answer to the above, I can re-introduce the use_qsatensmean code so that we have it  in case we need it. This PR would  then reduce to adding the code to write out the ensemble mean when use_qsatensmean is false.

jswhit:
@ClaraDraper-NOAA since you have verified that use_qsatmean=T/pseudo_RH=T is the same as pseudo_RH=F I'm OK with removing that option (with the caveat that global_workflow will need to be updated).  Sounds like the issue of how to renormalize q might be rather involved, and is best done separately.

src/enkf/params.f90
```diff
@@ -681,10 +675,6 @@ subroutine read_namelist()
        letkf_flag) then
      print *,'warning: no time localization in LETKF!'
    endif
-   if ((write_ensmean .and. pseudo_rh) .and. .not. use_qsatensmean) then
-      print *,'write_ensmean=T requires use_qsatensmean=T when pseudo_rh=T'
-      call stop2(19)
-   endif
```
jswhit, line 687: Do we need to raise an error and stop if write_ensmean=T and pseudo_RH=T (just remove the .and. not. use_qsatensmean part)?

src/enkf/controlvec.f90
```diff
@@ -229,56 +229,18 @@ subroutine read_control()
    end if
    !print *,'min/max qsat',nanal,'=',minval(qsat),maxval(qsat)
    q_ind = getindex(cvars3d, 'q')
-   if (use_qsatensmean .and. q_ind>0 ) then
-       allocate(qsatmean(npts,nlevs,nbackgrounds))
```
jswhit, line 233: Can also remove the variable declaration for qsatmean and qsat_tmp

ClaraDraper-NOAA:
> @ClaraDraper-NOAA since you have verified that use_qsatmean=T/pseudo_RH=T is the same as pseudo_RH=F I'm OK with removing that option (with the caveat that global_workflow will need to be updated). Sounds like the issue of how to renormalize q might be rather involved, and is best done separately.

Yes, good point.Are you OK with merging this as it is then?

jswhit:
> > @ClaraDraper-NOAA since you have verified that use_qsatmean=T/pseudo_RH=T is the same as pseudo_RH=F I'm OK with removing that option (with the caveat that global_workflow will need to be updated). Sounds like the issue of how to renormalize q might be rather involved, and is best done separately.
> 
> Yes, good point.Are you OK with merging this as it is then?

started a review with a couple of comments

jswhit:
May also have to revise regression tests if they have use_qsatmean=T

ClaraDraper-NOAA:
I've removed use_qsatensmean from the regression tests, and switched pseudo_RH from T to F - so all tests pass. Switching to the recommended pseudo_RH= True would fail the current tests, and require a new baseline be generated.

ClaraDraper-NOAA:
@CoryMartin-NOAA @jswhit Are there any more issues that I should address for this? If not, can you both please approve it?

jswhit:
@ClaraDraper-NOAA did you see my review comments above?

CoryMartin-NOAA:
@ClaraDraper-NOAA I am running a few days of cycles in our workflow just to confirm things are working as expected. I should be able to see results today or tomorrow and will approve it then. Thanks for making these changes.

CoryMartin-NOAA:
@ClaraDraper-NOAA just to clarify, if I run with PSEUDO_RH=.false. in this branch, and .true. in develop, with USE_QSATENSMEAN=.true., the results should be identical, or are they expected to be different? I ask because I am seeing changes in the q ensemble mean increment.

ClaraDraper-NOAA:
> @ClaraDraper-NOAA just to clarify, if I run with PSEUDO_RH=.false. in this branch, and .true. in develop, with USE_QSATENSMEAN=.true., the results should be identical, or are they expected to be different? I ask because I am seeing changes in the q ensemble mean increment.

Yes, I'd expect those to the same. Are you using the write ensemble mean option within the GSI EnKF, or using a different program to calculate the mean of the ensemble members after they're written out?

CoryMartin-NOAA:
@ClaraDraper-NOAA disregard, it is because the workflow checks out a GSI version by default that does not include https://github.com/NOAA-EMC/GSI/pull/568 yet. This tracer clipping difference is likely the cause, but let me verify.

src/enkf/params.f90
```diff
@@ -681,10 +675,6 @@ subroutine read_namelist()
        letkf_flag) then
      print *,'warning: no time localization in LETKF!'
    endif
-   if ((write_ensmean .and. pseudo_rh) .and. .not. use_qsatensmean) then
```
ClaraDraper-NOAA, line 684: @jswhit Responding to "Do we need to raise on error and stop if write_ensmean=T and pseudo_RH=T?". 

No, I moved the re-normalization by qsat to the top of the writecontrol routine, so it's now OK to use write_ensmean=T and psuedo_RH=T. Note: I got slightly different results when doing the normalization here then writing out the mean, than when I calculated the mean from the ensemble q/qsat output.

ClaraDraper-NOAA:
Pushed a change with the qsatmean and qsat_tmp declarations removed at Jeff's request.

src/enkf/params.f90
jswhit: OK thanks

src/enkf/params.f90
jswhit: Done

jswhit (approved):

jswhit2 (approved):

CoryMartin-NOAA (approved):
I ran some low resolution cycling test to make sure everything is working as intended, looks good to me!

PR #617, opened 2023-08-31
Error in processing VAD winds.

### LINKED ISSUE #615: Bug in read_prepbufr.f90
It was noted that there were a large number of observations that were being eliminated because there was no prepbufr observation type number in the setupw routine.  This should not occur because all of wind observations going into this routine should have a prepbufr observation type number.  

After considerable testing and printing values, it was found that the output array from the read_prepbufr routine had lines in it that contain only zeros.  The reason for this is that there is a section of code added to read_prepbufr that modifies the type 224 and newvad winds that was added after the output location (iout)was determined.  Since this code will jump out of the observation loop under certain conditions, those lines (a value of iout) of the output array will not be given values and will use the default values (zero).  For the test case I was using, this resulted in thousands of lines of zeros.

The solution is simple, move the processing of the type 224 and newvad data to before the values of iout are determined.

Note that duplicate lines in setupw to check for this condition exist and one set of checks can be removed (one check remains).

Doing this will result in a smaller array being passed to the observation distribution routines and fewer obs read into the setupw routines.  Fewer observations will also be put on the processor(s) that contain the 0,0 lat,lon location.  In my case 2924 wind (not real) observations were removed from 2 processors.  

The output in stdout will be changed with fewer obs on a few processors and the printout from setupw that removes these obs will be eliminated.

Since the bad obs are removed before being used, the results should not be changed.

In my case, the removal of these (not real) observations and the removal of associated code resulted in a wall time reduction from 931s to 870s.  Of course this is just one comparison, but the 931s was consistent with other runs with the same set of data.  None of the other runs completed in a time close to 889s.

All regression tests were passed.

Code is in my branch vadbug.

jderber-NOAA:
It was noted that a large number of observations were being eliminated by the setupw routine because there was no observation type.  This observation type check was clearly added because of previous problems with this data.  

It was found that there was an issue in the read_prepbufr routine because code for vad winds was added in the wrong location.  This resulted in many lines containing only zeros to the output file.

The code for processing the vad was moved before the determination of the output line to fix the issue.  It was necessary to move a few more lines to make everything work properly.  Also one of the checks (there were multiple checks) for a zero observation type was removed from setupw.  A check remains, but this condition should not occur unless another error is introduced.

This change results in the removal of many unrealistic observations from the output uv file from prepbufr (when vad winds are used).  Because of the existing observation type check analysis results are unchanged.

NOTE:  An additional change was made to allow the multiscale ensemble regional test I was using to work (both for the control and modified code).  In constants.f90 and gsi_rfv3io_mod.f90 a change was made to allow longer file names to be used for the input ensemble files.  

These changes address fixes #615 

- [ ] Bug fix (non-breaking change which fixes an issue)

These changes have been tested using the regression tests and a single regional case that is using vad winds.  All tests are passed with the only differences being in the output number of wind observation after distribution (the processors containing point 0,0 contained a lot of extra observations from the bug) and the output from the removal of the data by the observation type check.

Note because of the removal of many unrealistic observations in the processing, there appears to be a noticeable reduction in run time when vad winds are used.

 
**Checklist**

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] Any dependent changes have been merged and published

**DUE DATE for this PR is 10/12/2023**. If this PR is not merged into `develop` by this date, the PR will be closed and returned to the developer.


src/gsi/setupw.f90
JacobCarley-NOAA, line 1: It looks like this was some old debugging code that was cleaned up (great!). But I think `num_bad_ikx` could probably also be removed from the declaration portion at the top of setupw.

JacobCarley-NOAA (changes requested):
I'm good with the changes pending one very minor comment.

RussTreadon-NOAA:
**WCOSS2 ctests**
Install `develop` at [`9e5aa09`](https://github.com/NOAA-EMC/GSI/commit/9e5aa09c0a31ec3f4f8352f2a6de0dbf487d0b0d) and `jderber-NOAA:vadbug` at [`335c4b2`](https://github.com/jderber-NOAA/GSI/commit/335c4b276f4554b084a4d735af1778522922cce3) on Cactus.  Run standard suite of 9 ctests with following results

```
russ.treadon@clogin08:/lfs/h2/emc/da/noscrub/russ.treadon/git/gsi/pr617/build> ctest -j 9
Test project /lfs/h2/emc/da/noscrub/russ.treadon/git/gsi/pr617/build
    Start 1: global_3dvar
    Start 2: global_4dvar
    Start 3: global_4denvar
    Start 4: hwrf_nmm_d2
    Start 5: hwrf_nmm_d3
    Start 6: rtma
    Start 7: rrfs_3denvar_glbens
    Start 8: netcdf_fv3_regional
    Start 9: global_enkf
1/9 Test #8: netcdf_fv3_regional ..............***Failed  484.51 sec
2/9 Test #9: global_enkf ......................   Passed  488.58 sec
3/9 Test #5: hwrf_nmm_d3 ......................   Passed  492.74 sec
4/9 Test #7: rrfs_3denvar_glbens ..............   Passed  605.94 sec
5/9 Test #4: hwrf_nmm_d2 ......................   Passed  607.86 sec
6/9 Test #6: rtma .............................   Passed  1089.33 sec
7/9 Test #3: global_4denvar ...................   Passed  1324.26 sec
8/9 Test #1: global_3dvar .....................   Passed  1442.52 sec
9/9 Test #2: global_4dvar .....................   Passed  1563.17 sec
```

The `netcdf_fv3_regional` failure is due to a timing scalability test
```
The case has Failed the scalability test.
The slope for the update (1.509785 seconds per node) is less than that for the control (1.542529 seconds per node).
```
The wall times for the _updat_ (`jderber-NOAA:vadbug`) and _contrl_ (`develop`) do not show any anomalous behavior
```
netcdf_fv3_regional_hiproc_contrl/stdout:The total amount of wall time                        = 63.881463
netcdf_fv3_regional_hiproc_updat/stdout:The total amount of wall time                        = 63.013349
netcdf_fv3_regional_loproc_contrl/stdout:The total amount of wall time                        = 65.423992
netcdf_fv3_regional_loproc_updat/stdout:The total amount of wall time                        = 64.221177
```
This is not a fatal fail.

Check the `nkeep` value in `fort.202`.   `nkeep` is the number of wind observations passed out of `read_prepbufr.f90` for processing by setup routines.   Find that the _updat_ code consistently reduces the number of observation kept by `read_prepbufr`

|     |  _contrl_ | _updat_ |
| -- | --------- | --------- |
| global_3dvar | 3168472 | 3147394 |
| global_4denvar | 3172060 | 3150982 |
|  netcdf_fv3_regional | 323180 | 321662 |
| rrfs_3denvar_glbens | 159146 | 157394 |

While the number of kept observations decreases in _updat_, the number of assimilated observations is the same as _contrl_.   This is consistent with the explanation of the bug and it's fix in this PR.

src/gsi/setupw.f90
jderber-NOAA: This variable is still used.  I removed one case of checking for zero values, but I left the second one just in case.

JacobCarley-NOAA: Ah - I see now thanks.

JacobCarley-NOAA (approved):
Looks good.

jderber-NOAA:
Hera ctests

Regression tests were run on Hera earlier.  The results appear to be essentially identical to those Russ ran on WCOSS.  Even the same test failed for timing reasons.   Also, I checked the nkeep values on Hera and they are again identical as would be expected.

src/gsi/read_prepbufr.f90
```diff
@@ -2003,6 +2003,85 @@ subroutine read_prepbufr(nread,ndata,nodata,infile,obstype,lunout,twindin,sis,&
 ! don't use MESONET psfc obs if  8th character of station id is "x")
               if( kx==188 .and. psob .and. sidchr(8)=='x' ) usage=r100
 
+!             Set inflate_error logical based on qm flag
```
TingLei-NOAA, line 2006: @jderber-NOAA  Would you let me know what purpose is for this part of changes. Maye, it is only some simple moving of blocks for uvob. But it is difficult for me to trace the changes.  will these changes be supposed to change the results ?  The current regession tests are enough to verify changes here? (the GSI would go through this part)

TingLei-NOAA: @jderber-NOAA I think if changes in read_prepbufr.f90 could be put in another PR, the changes for processing VAD could be rather quickly add to the GSI repository through this PR.

jderber-NOAA: The change here is the important change to resolve the issue.  The processing of the VAD winds needs the surface elevation stuff so it has to be moved along with the actual VAD wind stuff.  As the code was, the VAD processing occurred after iout was determined.  Since the VAD wind processing jumps out of the loop sometimes, when it jumps out of the k loop it results in the cdata_ array being all zero for that value of iout.

jderber-NOAA: By the way, the results will not change.

TingLei-NOAA: Thanks for this explanation!

TingLei-NOAA (approved):

RussTreadon-NOAA (approved):
Two peer reviewers have looked over changes and approved.  WCOSS2 (Cactus) and Hera ctests pass with expected results. 

Approve.

PR #639, opened 2023-10-06
Add Regression Test for HAFS v1

JingCheng-NOAA:
**Description**
This PR will add four sets of regression tests for the current suite of HAFSv1 GSI. 
case 1: 3DEnvar with GDAS Ensemble plus FGAT capability.
case 2: 4Denvar with GDAS Ensemble plus FGAT capability.
case 3: Self-cycled 3Envar with HAFS Ensemble.
case 4: Self-cycled 3Envar with GDAS Ensemble plus HAFS Ensemble.

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)


hu5970:
@JingCheng-NOAA  This PR is not based on the latest develop branch. It does not have scripts for adding new ctest cases. Could you make it an issue instead of PR? Thanks, Ming

JingCheng-NOAA:
If makes it an issue, then the existing issue#600 is already about the
regression tests. I will remove this newly opened PR then.

Issue #678, opened 2024-01-08
AMSR2 Retrieval Predictor Caculation Went Wrong


XuLu-NOAA:
When calculating the predictors for AMSR2 retrievals in clw_mod.f90 between L987 and 988, Tb obs from channels 7 and 11 are always smaller than the corresponding channels 8 and 12.  This will result in negative values inside log and fail the debug mode. The test case is on 202008250600 UTC.
Need to contact the code provider or the data provider to understand if this is a change in the data nature or a bug. @MichaelLueken 

git blame  -L 987,988 clw_mod.f90
a02fd30b src/clw_mod.f90 (Michael.Lueken 2016-08-31 17:36:25 +0000    pred_var_clw(1) = log(tb_use(7) - tb_use(8))      !(tb18v - tb18h)
a02fd30b src/clw_mod.f90 (Michael.Lueken 2016-08-31 17:36:25 +0000    pred_var_clw(2) = log(tb_use(11) - tb_use(12))    !(tb36v - tb36h)

erinjones2:
@XuLu-NOAA @MichaelLueken If I had to guess, I would say this has something to do with the (apparently unsettled) question of polarization order for channels in the AMSR2 BUFR.

erinjones2:
@XuLu-NOAA @MichaelLueken Could this have anything to do with any lingering uncertainties with the AMSR2 BUFR files?

ADCollard:
@XuLu-NOAA Can you put this in some context, please.  We do not use amsr2 operationally in GSI as the correct re-mapped bufr file needed is not available.  @emilyhcliu and @xincjin-NOAA are looking into making this work.

TingLei-NOAA:
An update:  Thanks to discussions here and off-line exchanges from  @ADCollard and @xincjin-NOAA , @XuLu-NOAA and his HAFS colleagues decided , since this data will not be used for HAFS in the foreseeable future, that it is better to add a condition test in the code and, hence, when this data is used and the issue is identified, GSI would abort with error message.  By dosing so, future developers who will use this data will know where to start a thorough investigation and give a fix.  To have a quick adding of  the above abortion with error message , after discussion between Xu and Ting,  it is decided to add it to https://github.com/NOAA-EMC/GSI/pull/681.     @XuLu-NOAA  Please have a look to see if it is as you expected.

TingLei-NOAA:
[...]

PR #670, opened 2023-12-14
Complete issue #607 (Optimization of multiscale regional runs)

jderber-NOAA:
**Description**

This issue was intended to speed up regional multiscale runs.  The initial parts of these changes were intended to speed up the initial processing of the observations.  Unfortunately, I only able to complete changes to the reading of the observations.

The changes include the addition of the deallocates for many of the allocates in the reading routines, a restructuring of the thinning process and the addition of some new options for input pmot values.  

The restructuring of the thinning process was to set flags for the data that was thinned and flags for data that was qc'ed in the read routine.  Then based on the input pmot value the data was compressed to only retain the desired data.  The pmot values are:

pmot = 0. Pass on all observations that fail the setup quality control, but not those that are thinned.
pmot = 1. Pass on all observations that fail the setup quality control and the obs that are thinned.
pmot = 2. Pass on only the observations that do not fail the quality control and are not thinned.
pmot = 3. Pass on only the observations that do not fail the quality control plus the obs that are thinned.

The pmot = 0 and 1 options are the same was in the read_satwnd routine before, but the change has been included for all routines that thin the input data using the convthin routines.  The intent was to allow an option to minimize the data that is passed on to the setup routines if desired.  This is similar to what is done for the global AVN runs as compared to the FNL runs, except is applicable to individual observation types.

The routines for which the above changes were made to are:

    read_dbz_nc.f90 (pmot_dbz hardwired to 2 in obsmod)
    read_fl_hdob.f90
    read_goesimgr_skycover.f90
    read_l2bufr_mod.f90
    read_prepbufr.f90
    read_radar.f90 (pmot_vr hardwired to 0 in obsmod)
    read_radar_wind_ascii.f90 (pmot_vr hardwired to 0 in obsmod)
    read_rapidscat.f90
    read_satmar.f90
    read_satwnd.f90
    read_sfcwnd.f90
    read_wcpbufr.f90

The values for pmot are defined as real numbers in the convinfo file (except where noted above).

In addition some other minor changes were made:

    1. Changes to parallelization of control2state routines.
    2. The combination of ensctl2state and ensctl2state_ad into a single ensctl2state module with the inclusion of a ensctl2state_set routine to define flags once. This is done in a way that is consistent with what is done with control2state above and allows the flags to be calculated once rather than repeatedly every time the routines are called.
    3. Deallocating arrays in read_* routines that were left allocated. (removed considerable memory)
    4. Making some of the muse flags in setup routines dependent on qc mark as well as usage flag.
[...]

PR #681, opened 2024-01-11
a quick fix for Issue:A indexing out of bounds issue shown in the global_4denvar regr…

### LINKED ISSUE #676: A indexing out of bounds issue shown in the global_4denvar regression test in setuprad.f90
When trying to run various regression tests for a PR using debug mode GSI, an issue within the current EMC GSI is found.  in the global_4denvar regression test,  there are error messages:
```
forrtl: severe (408): fort: (2): Subscript #1 of the array ABI2KM_BC has value 4 which is greater than the upper bound of 3

Image              PC                Routine            Line        Source
gsi.x              0000000007AA763F  Unknown               Unknown  Unknown
gsi.x              00000000054E656A  rad_setup_mp_setu        1108  setuprad.f90
gsi.x              0000000003F6488C  gsi_radoper_mp_se         100  gsi_radOper.F90
gsi.x              00000000025FC8CC  setuprhsall_              491  setuprhsall.f90
....
```
The L1108 in setuprad.f90 is 
```
      abi2km_bc = zero
       abi2km_bc(2) = 233.5_r_kind
        abi2km_bc(3) = 241.7_r_kind
        abi2km_bc(4) = 250.5_r_kind   !L1108
      ````
      The abi2km_bc is declared as "real(r_kind),dimension(nchanl) :: abi2km_bc".  
      so seems when nchanl is less than 4, the error would occur as in this regression test. 
A quick fix for considering the actual value of nchanl  is needed

### PR DESCRIPTION
**Pull request by TingLei-daprediction**
In response to Issue:" A indexing out of bounds issue shown in the global_4denvar regression test in setuprad.f90 #676" (https://github.com/NOAA-EMC/GSI/issues/676)  and following both online and off-line discussions, it is decided to retain the abi2km part in setuprad.f90 ( for potential future development/improvement), while a simple index-related checks are added to prevent the indexing error when GSI is built with debug mode. 
It is important to note that  this indexing error does not impact GSI results (when GSI is built with optimization options) since abi2km is currently hardwire consfigured to be not used. 

Fixes #676 

**DUE DATE for merger of this PR into `develop` is 2/22/2024 (six weeks after PR creation).**

### CODE REVIEW

**Review by XuLu-NOAA — COMMENTED** [review]
Hi, @TingLei-daprediction , is there a convention that the variables stay on the left side of the numbers in equation/condition? E.g. size(abi2km_bc) .ge. 2 instead of 2 .le. size(abi2km_bc).  Other than that I don't have questions about the fix itself.

**Comment by TingLei-daprediction** [comment]
@XuLu-NOAA  when I have those lines,  kind of like I first have "if (i > size (abi2km_bc) ) then ",  where I replaced i with 2 , 3 and so on.   let me know if you prefer the form as "size(abi2km_bc) > 2" and so on.

**Comment by RussTreadon-NOAA** [comment]
As @wx20jjung notes we need understand the reason for the `abi2km` code.   

Logical `abi2km` is local to `setuprad.f90` and is initialized to `.false.`
```
  abi        = obstype == 'abi'
  abi2km     = .false.

  ssmis=ssmis_las.or.ssmis_uas.or.ssmis_img.or.ssmis_env.or.ssmis
```
I do not see anywhere in `src/gsi` where `abi2km` is set to `.true.`.  I don't even see a mechanism by which `abi2km` can be set to `.true.` apart from changing the hardcoded value in `setuprad.f90`.  

Since `abi2km` is hardwired to `.false.`, the  code block in which `abi2km_bc` is used, namely,
```
           if (abi2km .and. regional) then                                                                                                                                  
              pred(:,i) = zero
              if (i>=2 .and. i<=4) then
                 if (tb_obs(i) > 190.0_r_kind .and. tb_obs(i) < 300.0_r_kind) then
                    pred(1,i)=1.0_r_kind
                    pred(2,i)=tb_obs(i)-abi2km_bc(i)
                    pred(3,i)=(tb_obs(i)-abi2km_bc(i))**2
                    pred(4,i)=(tb_obs(i)-abi2km_bc(i))**3
                 end if
              end if
           end if
```
will never be executed.   Hence Jim's question.  Is the `abi2km` code obsolete, orphaned, or in need of a rewrite?

If we are not ready to remove or rewrite code related to `abi2km` and `abi2km_bc`, we can restrict the initialization of `abi2km_bc` to when `abi2km` and `regional` are true.  That is, modify the initialization of `abi2km_bc` to
```
        if (abi2km .and. regional) then
           abi2km_bc = zero
           abi2km_bc(2) = 233.5_r_kind
           abi2km_bc(3) = 241.7_r_kind
           abi2km_bc(4) = 250.5_r_kind
        endif
```
Since `abi2km` is `.false.`, neither this block nor the block which uses `abi2km_bc` will be entered in regional or global gsi runs.

**Comment by TingLei-daprediction** [comment]
@RussTreadon-NOAA . I can't answer the question if ab2km is obsolete or it will be updated/refreshed in the future.  That is why I retain them and give the current fix just to avoid indexing errors which would cause GSI abort  when GSI is built with debug mode. 
Second, if we use the conditions (abi2km.and.regional) , this would also prevent the abortion as reported for the global regression test as reported in the corresponding issue and leave further sorting out /treatment to the future if abi2km would be used in the future.   Do you prefer this fix?

**Comment by JingCheng-NOAA** [comment]
This piece of code related to `abi2km` seems like someone's testing code to me. But in case it was designed for future use, I agree we can block this part for now. And I am more toward to @RussTreadon-NOAA 's solution to block this part at the initialization part. I have no more questions related to this fix.

**Comment by JingCheng-NOAA** [comment]
> @RussTreadon-NOAA . I can't answer the question if ab2km is obsolete or it will be updated/refreshed in the future. That is why I retain them and give the current fix just to avoid indexing errors which would cause GSI abort when GSI is built with debug mode. Second, if we use the conditions (abi2km.and.regional) , this would also prevent the abortion as reported for the global regression test as reported in the corresponding issue and leave further sorting out /treatment to the future if abi2km would be used in the future. Do you prefer this fix?

@TingLei-daprediction then how about remove the `regional` in the condition, only use `abi2km`?

**Comment by TingLei-daprediction** [comment]
@JingCheng-NOAA  I have no preference on this!

**Comment by XuLu-NOAA** [comment]
Can we do both? The if condition will ensure the initialization is not used by default. But index check also resolves one potential issue if the abi2km is turned on in the future. Initialized or not, it is still a bug.

**Comment by RussTreadon-NOAA** [comment]
My preference is to entirely remove the abi2km code because it won't be executed given how it is currently implemented in `setuprad.f90`.   

GSI source code is version controlled.  Removing abi2km code does not remove it from the repository.   A developer interested in the abi2km code can locate the PR at which it was removed via a search of github issues or PRs and, given this, checkout the GSI hash prior to the commit at which the code was removed.

**Comment by TingLei-daprediction** [comment]
@XuLu-NOAA  and @JingCheng-NOAA   So, since @RussTreadon-NOAA  has the same opinion as  @wx20jjung, one of the developers for this part, that this abc2km could be removed,  I have removed abc2km blocks and please review the current version.   Thanks.

**Review by XuLu-NOAA — APPROVED** [review]

**Review by ShunLiu-NOAA — DISMISSED** [review]

**Comment by ShunLiu-NOAA** [comment]
@JingCheng-NOAA  please review the new changes from Ting.

**Review by RussTreadon-NOAA — DISMISSED** [review]
Thank you, @TingLei-daprediction , for removing unused code from `setuprad.f90`

Approve.

**Comment by JingCheng-NOAA** [comment]
I have no more issues with the new changes.

**Review by JingCheng-NOAA — APPROVED** [review]

**Comment by ShunLiu-NOAA** [comment]
@TingLei-daprediction Could you please run regression test on HERA, WCOSS and ORION?

**Comment by TingLei-NOAA** [comment]
@ShunLiu-NOAA  An update on the regression tests:  on hera, all passed, on orion, netcdf_fv3_regional failed for :
```
The memory for netcdf_fv3_regional_loproc_updat is 303768 KBs.  This has exceeded maximum allowable memory of 278040 KBs,
resulting in Failure memthresh of the regression test.
```
Considering this PR's property, it is safe to say this memory increase has nothing to with GSI itself and this regression test could be regarded as "pass" in my opinion.     The regression test results on wcoss2 are to be reported when it is available to me again.

**Review by XuLu-NOAA — APPROVED** [review]
Thanks for adding these comments!

**Review by JingCheng-NOAA — APPROVED** [review]

File: src/gsi/clw_mod.f90 [hunk 1]
```diff
@@ -983,6 +983,12 @@ subroutine retrieval_amsr2(tb,nchanl,clw,kraintype,ierr)
 
 
 ! ---------- Calculate predictors ---------------------------
+  if(.not.( (tb_use(7) - tb_use(8)) > 0_r_kind ).or. &
+     .not.( (tb_use(11) -tb_use(12)) > 0_r_kind ) ) then
+     write(6,*)'The issue (GSI#678) needs further investigation and be resolved if this kind of obs is used' 
+     write(6,*)'Now, GSI would abort'
+     call stop2(72)
```
RussTreadon-NOAA: While I understand the rationale for this change (detect a protentional error), we do not want to merge this logic into `develop`.   

There are other places in `clw_mod` where a difference of two terms is used as the argument of `log`. Logic is in place to ensure the sign of the difference is greater than zero. 
 For example, lines 801 to 808 of `clw_mod.f90` in `develop` read
```
! ---------- Calculate predictors ---------------------------                                      

  if(tb_use(4) > tb_use(3)) then
     return
  endif

  pred_var_clw(1) = log(tb_use(3)-tb_use(4))  !(tb18v - tb18h)                                     
  pred_var_clw(2) = log(tb_use(6)-tb_use(7))   !(tb37v - tb37h)                                    
```
While this does not check for the possibility of a negative argument on the second `log`,  The first difference is checked.  This may be sufficient.  I am not familiar the characteristics of GMI channels.

Similar code could be placed before lines 993 and 994 and the above logic, write, and stop removed.

TingLei-daprediction: @RussTreadon-NOAA  As some explanation at https://github.com/NOAA-EMC/GSI/issues/678, the problem is that we need more significant efforts to investigate the issue and the solution. For example, we don't know if 
```
if(tb_use(4) > tb_use(3)) then
     return
  endif
  ```  ( return to avoid the exception) is the one needed. It is not preferable because it won't give an warnming to the user/developer who might still think everything is going well.  That is why it is considered the current logics ( error message and stop) is needed before , in the future, when the data is to be considered and further work is applied on this issue.    That will be great if you share your opinions on that issue and more experts could discuss to decide a fix for being now.

RussTreadon-NOAA: @TingLei-daprediction .  A similar argument could be made for other checks on the `log` argument found in `clw_mod.f90`.  These checks return execution or bypass `log` of an invalid value without writing to stdout and aborting `gsi.x` execution.  Yes, this means the invalid `log` argument goes undetected.   I assume the developers who wrote the code understood this and are OK with the given behavior.

The current write & abort change to `clw_mod.f90` in this PR is OK for testing but not for merger into `develop`.  If we want to add a check, I suggest mimicking the logic found in a similar section of `retrieval_gmi`.  

As @ADCollard explained we are not operationally processing AMSR2 data in the GSI.  @emilyhcliu and @xincjin-NOAA are looking into this.  I take this to mean that the GSI should not process AMSR2 data for the time being.

emilyhcliu: @RussTreadon-NOAA You are correct.  We do not have the correct AMSR-2 BUFR from NESDIS yet.  In the meantime, we are using a few sample AMSR-2 BUFR data from GMAO for testing AMSR-2 in GSI.

TingLei-NOAA: @RussTreadon-NOAA  "I assume the developers who wrote the code understood this and are OK with the given behavior." I am not sure if I understand what you meant.   The codes is always improving. It is obvious the "developer who wrote the code" would not be ok with the use of negative values in log function  in their  code(while it  exists to be deal with through this issue/PR) .

RussTreadon-NOAA: @TingLei-daprediction , my comments are not meant to stop merger of this PR into GSI `develop`.  As stated above, I am not in favor of merging this type of print & abort into the authoritative GSI `develop`.  I am neither the GSI code manager nor the Handling Reviewer for this PR.  My comments do not preclude merger.

If the team decides to retain the proposed change to `clw_mod.f90` in this PR, please update the PR description to note that this PR resolves issue #678 in addition to resolving #676.

xincjin-NOAA: I agree with @RussTreadon-NOAA's opinion.  At least GSI shouldn't be stopped in this situation because in the worst situation this type of satellite data, e.g., AMSR-2,  is not been used but there are many other Satellite data are still useful.  There is also another situation in which only a few points that are not valid  in the dataset. 

Print out message in this situation might also be not acceptable. Because this will print out tons of outputs saying one thing.

TingLei-daprediction:
@RussTreadon-NOAA  I had removed this change in question from this PR and will document our discussion in the issue 678 and let the decision be made/documented there.   Thanks.

RussTreadon-NOAA:
Thank you @TingLei-daprediction

TingLei-NOAA:
An update:  the regressions tests on WCOSS2 all passed!

ShunLiu-NOAA (dismissed):

RussTreadon-NOAA:
@ShunLiu-NOAA and @TingLei-daprediction 

With the merger of PR #670 into `develop` the changes to `setuprad.f90` in this PR conflict with the updated `develop`.  Resolution of the conflict is straight forward.   See Cactus `/lfs/h2/emc/da/noscrub/russ.treadon/git/gsi/pr681` for a working copy of branch `feature/gsi_abi2km_index` with the conflict to `setuprad.f90` resolved.

@TingLei-NOAA , if you like I can attempt to push my updated working copy of `feature/gsi_abi2km_index` to your fork.  This may not work. I don't know.  I won't try unless you approve.

TingLei-NOAA:
@RussTreadon-NOAA  Thanks for taking care of this conflict.  I had sent the invitation to you as the collaborator on my fork and hence, after you accept it,  you could push your change to this PR.

RussTreadon-NOAA:
Conflict resolved @ [35b8aaf](https://github.com/NOAA-EMC/GSI/commit/35b8aaf15f371e7e204747bd4b54d5e98bf0aca2).

@TingLei-NOAA , peer reviewers need to re-approve this PR along with @ShunLiu-NOAA , the handling reviewer.

TingLei-NOAA:
@XuLu-NOAA @JingCheng-NOAA  your another round of reviewing for the updated PR will be appreciated.

JingCheng-NOAA (approved):

XuLu-NOAA (approved):

ShunLiu-NOAA:
@TingLei-daprediction Could you please re-run regression test?

TingLei-NOAA:
@ShunLiu-NOAA  All tests passed the reproducibility  tests on hera, orion and wcoss2.
On each machine, a few tests failed for either exceeding time or memory , which are not considered issues for this PR. 
In conclusion,  this PR passes all regression tests on the above 3 machines.

ShunLiu-NOAA (approved):

ShunLiu-NOAA:
@TingLei-NOAA Thank you for completing regression test on WCOSS2, HERA and ORION.

PR #940, opened 2025-10-15
Fix a bug in GNSS-RO assimilation with multiple BUFR observation files

### LINKED ISSUE #939: Unexpected observation error inflation in GNSS-RO assimilation with multiple BUFR observation files
In Global Navigation Satellite System (GNSS)-Radio Occultation (RO) assimilation, observation data are assigned a unique ID for each profile, which is used for inflation of observation errors (superobing at vertical levels). However, when multiple RO BUFR observation files are inputted, duplicated profile IDs can occur across data from different BUFR files, leading to an unexpected (over-inflated) result of observation error inflation.

Note that this issue matters only when multiple RO BUFR observation files are used. For a case with only a single RO BUFR input file, such as in the NCEP GFS operation, this should not be a problem. However, if one considers doing experiments with multiple RO BUFR input files, this issue needs to be addressed.

__Proposed solution:__ Ensure that RO profile IDs remain unique even when multiple RO BUFR observation files are inputted. This change may be made in the 'obs_para' subroutine.

gylien:
**Description**

This addresses issue #939 .
Resolves #939

In Global Navigation Satellite System (GNSS)-Radio Occultation (RO) assimilation, observation data are assigned a unique ID for each profile, which is used for inflation of observation errors (superobing at vertical levels). However, when multiple RO BUFR observation files are inputted, duplicated profile IDs can occur across data from different BUFR files, leading to an unexpected (over-inflated) result of observation error inflation.

Note that this issue matters only when multiple RO BUFR observation files are used. For a case with only a single RO BUFR input file, such as in the NCEP GFS operation, this should not be a problem. However, if one considers doing experiments with multiple RO BUFR input files, this issue needs to be addressed.

The revised code re-computes the RO profile ID in the 'obs_para' subroutine to ensure its uniqueness and thus fix the problem.

**Type of change**

- [x] Bug fix (non-breaking change which fixes an issue)

**How Has This Been Tested?**

This change has been tested and used in an operational global forecast system at the Taiwan Central Weather Administration (CWA) since 2024. The assimilation result should be the same when using a single RO BUFR input file; with multiple RO BUFR input files, the erroneous (overly large) observation error inflation is corrected.

The authors do not have access to any NOAA platform.
  
**Checklist**

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] Any dependent changes have been merged and published


XuanliLi-NOAA:
@gylien Thanks for updating GSI to support multiple GPSRO BUFR files. Could you share a bit more background on why multiple input GPSRO BUFR files are needed? Is it mainly for research purposes or for operational considerations? Also, are any changes required in read_gps.f90 or other modules to enable reading multiple GPSRO BUFR files?

gylien:
@XuanliLi-NOAA At CWA, we assimilate COSMIC-2 RO data processed by our local data processing center, Taiwan Analysis Center for COSMIC (TACC), (1st BUFR file) and assimilate other RO data processed by CDAAC available on NOMADS (2st BUFR file). Therefore, we input two RO BUFR files into our operational GFS/GSI system.

Even without any code modification, GSI can read multiple RO BUFR files and assimilate all RO data smoothly (by specifying multiple RO BUFR files in the namelist). The only problem is that the profile IDs assigned to each RO profile become not unique. For example, the profiles in the 1st BUFR file are assigned ID = 1, 2, 3, …, and the profiles in the 2nd BUFR file are again assigned ID = 1, 2, 3, …. Then, when GSI computes observation error inflation based the number of RO data within each model vertical level (in my understanding, equivalent to superobing at vertical levels), it mistakenly treats the two profiles having the same profile ID as the same profile. As a result, the counts are overestimated, which leads to erroneously large observation error inflation and too small observation impacts for RO data in both BUFR files. This happens without any warning.

In fact, we had assimilated the RO data operationally in this way (using two BUFR files) for years until we noticed this issue. After that, we fixed the issue by the current method: modifying the `obs_para` subroutine in `src/obs_para.f90` to ensure the uniqueness of RO profile IDs. There is no need to modify other modules such ad `read_gps.f90` to enable reading multiple RO BUFR files.

Of course, one can always concatenate all RO data from multiple sources into a single BUFR file to avoid this problem. I think that if NOAA does not anticipate any potential case that multiple RO BUFR files are used in GSI assimilation, this update may not be necessary.

However, if you are interested in accepting this update, please note that we made this fix based on a rather old version of the GSI code (as you can find in the commit). Although I already merged the latest develop branch with this fix, I have no chance to test the merged code. You may wish to test it on a NOAA platform to confirm that:
1. When only a single RO BUFR file is used, the assimilation results with and without this fix are identical;
2. When multiple RO BUFR files are used, the assimilation results with and without this fix are different, and the result with this fix should be close to that by concatenating the multiple RO BUFR files into a single file and assimilating the single file.

Thank you for your time to this issue.

XuanliLi-NOAA:
@gylien I see — you’re likely doing so by adding additional RO bufr filenames in obs_input section. 
The issue seems to be that when two or more RO BUFR files are used, read_gps assigns a list of profile IDs for each individual file, and these IDs can overlap. In genstats, the observation error is then inflated by the number of obs within a layer from the same profile — which may now include data from multiple profiles. As a result, the calculated error becomes larger than it should be.

At EMC, we typically use a single GPSRO BUFR file, or we concatenate multiple RO files into one for testing and development. Because of this, we haven’t encountered the issue you described. We’ll likely continue using this approach in the near future.

Thank you so much for bringing this to our attention and for providing a solution to the problem. I also really appreciate your detailed explanation — it was very helpful.

gylien:
@XuanliLi-NOAA Yes, what you described was exactly the problem. Thanks!

RussTreadon-NOAA:
Thank you @gylien for opening this PR.

Please note the following from the [GSI wiki](https://github.com/NOAA-EMC/GSI/wiki)

> The Unified Forecast System (UFS) will use the Joint Effort for Data assimilation Integration (JEDI) for its DA infrastructure. JEDI components, as they mature, will incrementally replace operational GSI-based components.

> Given this, the focus of the [NOAA-EMC/GSI repository](https://github.com/NOAA-EMC/GSI) has shifted to operational support during this transition. NOAA-EMC/GSI only supports current and planned operational NWS realizations of the GSI. Changes to NOAA-EMC/GSI must have a clear path to implementation with agreement from operational GSI application leads.

We need to consult with EMC management, subject matter experts, application leads to see if the changes in this PR are required for current or planned operational realizations of the GSI.

RussTreadon-NOAA:
**WCOSS2 ctests**

Install `CWA-NWP:feature/ro-multi-files` at 6b39782 as _updat_ and `develop` at d471712 as _contrl_ on Cactus.  Run GSI ctests with the following results
```
Test project /lfs/h2/emc/da/noscrub/russ.treadon/git/gsi/pr940/build
    Start 1: global_4denvar
    Start 2: rtma
    Start 3: rrfs_3denvar_rdasens
    Start 4: hafs_4denvar_glbens
    Start 5: hafs_3denvar_hybens
    Start 6: global_enkf
1/6 Test #3: rrfs_3denvar_rdasens .............   Passed  2064.31 sec
2/6 Test #2: rtma .............................   Passed  2372.88 sec
3/6 Test #6: global_enkf ......................   Passed  2431.81 sec
4/6 Test #5: hafs_3denvar_hybens ..............   Passed  2495.83 sec
5/6 Test #4: hafs_4denvar_glbens ..............   Passed  3086.28 sec
6/6 Test #1: global_4denvar ...................   Passed  3436.07 sec

100% tests passed, 0 tests failed out of 6

Total Test time (real) = 3440.42 sec
```

This result is expected.  The changes in this PR do not impact the gnss-ro processing using a single input bufr file.

PR #750, opened 2024-06-03
Update RRFS regression test cases

hu5970:
**Description**

1) Change RRFS case from rrfs_3denvar_glbens to rrfs_3denvar_rdasens.
       The new RRFS case has the same configuration as RRFS_A warm cycles.
2) delete case netcdf_fv3_regional
3) add RRFS EnKF case for conventional observations: rrfs_enkf_conv

This PR depends on GSI-fix PR: https://github.com/NOAA-EMC/GSI-fix/pull/20
This pull request is addressing Fix #586 

Please copy the data for the new cases:
Hera: "/scratch1/BMC/wrfruc/mhu/code/data"
Cactus:"/lfs/h2/emc/ptmp/Ming.Hu/data"

**Type of change**

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

**How Has This Been Tested?**

Those two new cases were tested on Hera and WCOSS2 (Cactus)

**Checklist**

- [x ] My code follows the style guidelines of this project
- [x ] I have performed a self-review of my own code
- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] New and existing tests pass with my changes
- [x ] Any dependent changes have been merged and published

Issue #323, opened 2022-03-02
nonphysical increments produced by serial filter on Hera


AndrewEichmann-NOAA:
Running the serial filter with EFSOI cycling (and also without at times) produces on some cycles non-physical increments, example following:

[Andrew.Eichmann@v71a1 atmos]$ grep increment gdas.t12z.enkfstat
[0] ens. mean anal. increment min/max  u   -107693.046875        29170.9941406    
[0] ens. mean anal. increment min/max  v   -45611.8164062        44759.7421875    
[0] ens. mean anal. increment min/max  tv   -9793.13574219        13133.6269531    
[0] ens. mean anal. increment min/max  q   -1219.31628418        6754.96484375    
[0] ens. mean anal. increment min/max  oz  -0.825550232548E-04   0.365831547242E-04
[0] ens. mean anal. increment min/max  ps   -1780.73510742        3868.44970703    
...

This is while running on WCOSS, and the result is the ensemble forecast crashing. In this run it is apparently so far the only time in 13 model days, to judge by grepping the remaining enkfstats produced with the serial filter. This particular problem (or this manifestation of it) has appeared before while running on Orion, after merging the GSI master of Jan 10 into the EFSOI fork, but then was resolved after merging from the Jan 25 hash of GSI master. Is there a standard approach to resolving this kind of problem?

The comrot and particular subdirectory is /gpfs/dell1/ptmp/Andrew.Eichmann/baily/efsoigdas.20211212/12/atmos (I've tarred the contents, minus the ensemble members, and left it on noscrub so it doesn't get purged; let me know if I should tar the members as well)
expdir: /gpfs/dell2/emc/modeling/save/Andrew.Eichmann/para/baily

Thank you!

AndrewEichmann-NOAA:
@jswhit2 Have you seen this come up before?

jswhit:
That's disturbing - no I haven't seen that before.

jswhit:
Is it reproducible (in other words does it happen again if you just run the whole cycle over again, including the GSI observer step?)

TingLei-NOAA:
[...]

PR #961, opened 2025-12-03
Resolve spuriously large EnKF increments by disabling shared-memory window in `mpi_readobs.f90`

### LINKED ISSUE #960: Spuriously large and unphysical increments in EnKF on `release/rrfs.v1.0.0`
When running EnKF retros in the `dev-sci` branch of rrfs-workflow, we found that the increments can be extremely large: 

Here are increments from the first cycle of a conus13km retro using 30 members and only assimilating aircar temperature and wind observations: 

```
ens. mean anal. increment min/max  u   -400.986419678        526.739135742    
ens. mean anal. increment min/max  v   -299.293304443        286.644195557    
ens. mean anal. increment min/max  t   -50.3178100586        261.687805176    
ens. mean anal. increment min/max  oz  -0.537249161425E-05   0.100919532997E-04
ens. mean anal. increment min/max  ps   -101.541290283        21.7783260345    
```

These increments are unphysical and crash the subsequent ensemble forecasts. 

Updates to `src/enkf/mpi_readobs.f90` can resolve this problem: 
```
ens. mean anal. increment min/max  u   -24.8986816406        27.3974285126    
ens. mean anal. increment min/max  v   -30.1263866425        35.6076431274    
ens. mean anal. increment min/max  t   -4.40757656097        6.25625514984    
ens. mean anal. increment min/max  oz  -0.423902207558E-06   0.442797130518E-06  
ens. mean anal. increment min/max  ps   -1.66400146484        2.14062714577   
```

Here is the ChatGPT summary of the changes made to `mpi_readobs.f90`: "this update fixes the large spurious increments because it removes a fragile shared memory plus “root only” reduction pattern and replaces it with a straightforward, globally consistent reduction of a normal Fortran array. That prevents corrupted or incomplete ensemble priors in observation space, which were manifesting as giant outlier perturbations and thus giant increments". However, we are still not sure if this problem is specifically due to GSI or something with the configuration in rrfs-workflow. We will need more testing once `mpi_readobs.f90` is updated.

SamuelDegelia-NOAA:
**Description**

As mentioned in #960, we find spuriously large increments in the EnKF when running RRFS retros with the `release/rrfs.v1.0.0` branch. 

This PR resolves the large increments by modifying `src/enkf/mpi_readobs.f90` to stop using a shared-memory window for `anal_ob`, instead allocating it as a standard local array on each rank. It also switches the MPI reductions from a node-level communicator to `MPI_COMM_WORLD` and adjusts which sections are guarded by the shared-memory root logic so that reductions occur on all ranks rather than only on the shared-memory root. 

Fixes #960

**Type of change**

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

**How Has This Been Tested?**

Tested by running RRFS retros on WCOSS2 for three cycles. Increment magnitudes were much more reasonable after this update. 
  
Note that we cannot run the ctests in the `release/rrfs.v1.0.0` or `release/rrfs.v1.1.0` branches since the regression tests lag behind the develop branch. 

**Checklist**

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] Any dependent changes have been merged and published

### DIFF
--- a/src/enkf/mpi_readobs.f90
+++ b/src/enkf/mpi_readobs.f90
@@ -33,7 +33,7 @@ module mpi_readobs
 !$$$
   
 use kinds, only: r_double,i_kind,r_kind,r_single,num_bytes_for_r_single
-use params, only: ntasks_io, nanals_per_iotask, nanal1, nanal2
+use params, only: ntasks_io, nanals_per_iotask, nanal1, nanal2, use_shmem_window
 use radinfo, only: npred
 use readconvobs
 use readsatobs
@@ -148,7 +148,11 @@ subroutine mpi_getobs(obspath, datestring, nobs_conv, nobs_oz, nobs_sat, nobs_to
     ! associate fortran pointer with c pointer to shared memory 
     ! segment (containing observation prior ensemble) on each task.
     call MPI_Win_shared_query(shm_win, 0, segment_size, disp_unit, anal_ob_cp, ierr)
-    allocate(anal_ob(nanals, nobs_tot))
+    if (use_shmem_window) then
+       call c_f_pointer(anal_ob_cp, anal_ob, [nanals, nobs_tot])
+    else
+       allocate(anal_ob(nanals, nobs_tot))
+    endif
     ! initialize shared memory window.
     anal_ob=0
     if (neigv > 0) then
@@ -249,27 +253,37 @@ subroutine mpi_getobs(obspath, datestring, nobs_conv, nobs_oz, nobs_sat, nobs_to
     if (nproc == 0) t1 = mpi_wtime()
 ! exchange obs prior ensemble members across all tasks to fully populate shared
 ! memory array pointer on each node.
+    if ((use_shmem_window .and. nproc_shm == 0) .or. (.not. use_shmem_window)) then
        if (real(nanals)*real(nobs_tot) < 2_r_kind**32/2_r_kind - 1_r_kind) then
-          call mpi_allreduce(mpi_in_place,anal_ob,nanals*nobs_tot,mpi_real4,mpi_sum,mpi_comm_world,ierr)
+          if (use_shmem_window) then
+             call mpi_allreduce(mpi_in_place,anal_ob,nanals*nobs_tot,mpi_real4,mpi_sum,mpi_comm_shmemroot,ierr)
+          else
+             call mpi_allreduce(mpi_in_place,anal_ob,nanals*nobs_tot,mpi_real4,mpi_sum,mpi_comm_world,ierr)
+          endif
        else
           ! count won't fit in 32-bit integer and mpi_allreduce doesn't handle
           ! 64 bit counts.  Split up into smaller chunks.
           mem_ob = 0.
           do na=1,nanals
               mem_ob(:) = anal_ob(na,:)
-              call mpi_allreduce(mpi_in_place,mem_ob,nobs_tot,mpi_real4,mpi_sum,mpi_comm_world,ierr)
+              if (use_shmem_window) then
+                 call mpi_allreduce(mpi_in_place,mem_ob,nobs_tot,mpi_real4,mpi_sum,mpi_comm_world,ierr)
+              else
+                 call mpi_allreduce(mpi_in_place,mem_ob,nobs_tot,mpi_real4,mpi_sum,mpi_comm_shmemroot,ierr)
+              endif
               anal_ob(na,:) = mem_ob(:)
           enddo
        endif
        !print *,nproc,'min/max anal_ob',minval(anal_ob),maxval(anal_ob)
-    if (nproc_shm == 0) then
-       if (neigv > 0) then
-          mem_ob_modens = 0.
-          do na=1,nanals
-             mem_ob_modens(:,:) = anal_ob_modens(neigv*(na-1)+1:neigv*na,:)
-             call mpi_allreduce(mpi_in_place,mem_ob_modens,neigv*nobs_tot,mpi_real4,mpi_sum,mpi_comm_shmemroot,ierr)
-             anal_ob_modens(neigv*(na-1)+1:neigv*na,:) = mem_ob_modens(:,:)
-          enddo
+       if (use_shmem_window .or. nproc_shm == 0) then
+          if (neigv > 0) then
+             mem_ob_modens = 0.
+             do na=1,nanals
+                mem_ob_modens(:,:) = anal_ob_modens(neigv*(na-1)+1:neigv*na,:)
+                call mpi_allreduce(mpi_in_place,mem_ob_modens,neigv*nobs_tot,mpi_real4,mpi_sum,mpi_comm_shmemroot,ierr)
+                anal_ob_modens(neigv*(na-1)+1:neigv*na,:) = mem_ob_modens(:,:)
+             enddo
+          endif
        endif
     endif
     if (nproc == 0) then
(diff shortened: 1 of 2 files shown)


SamuelDegelia-NOAA:
It looks like I do not have access to request reviewers here. But I will tag @ShunLiu-NOAA and @HuiLiu-NOAA for possible review.

HuiLiu-NOAA (approved):
The change works.

RussTreadon-NOAA:
It would be a good idea to invite @jswhit to review these changes.  Jeff is the lead developer of the GSI-EnKF.

ShunLiu-NOAA:
@RussTreadon-NOAA, Ming, Hui, and Sam encountered this issue during RRFS v1 testing. I'm not sure if Global GSI has seen something similar.

RussTreadon-NOAA:
@AndrewEichmann-NOAA reported unphysical enkf increments when running the serial filter on Hera (see issue #323)  I don't recall seeing unphysical increments on WCOSS2 ... but I don't routinely monitor this.

Tagging @CatherineThomas-NOAA for awareness

jswhit:
The shared memory segment is used to save memory when more than one mpi task is running on each node.  This change will substantially increase the memory footprint of the EnKF (especially for the LETKF which is currently run in operations).  Perhaps it would be better to figure out why the shared memory window is causing unphysical increments?  (or add a namelist parameter to optionally disable use of the shared memory window)

SamuelDegelia-NOAA:
Here are the memory usages for some additional tests with and without this update to `mpi_readobs.f90`.  

Without the change: 
`The maximum resident set size (KB)                   = 1124532`

With the change: 
`The maximum resident set size (KB)                   = 1130328`

Assuming that this is the correct line to look at, the memory impact seems relatively small (~0.5% increase). However, this comparison is for the serial filter and not the LETKF (which @jswhit mentions will have a larger impact).

I should note though that this PR is specifically for the RRFS branch for which we only plan to use the serial filter.

jswhit:
In order to get this working for you while we investigate the underlying cause, I'm OK merging this change as long as it is made optional (and not the default).  This could be done by introducing a `use_shmem_window` logical namelist variable.

SamuelDegelia-NOAA:
Thanks for your suggestion @jswhit. I just added a namelist option `use_shmem_window` that defaults to true and uses the original shared-memory logic in `mpi_readobs.f90`. If we set `use_shmem_window=.false.`, then the code uses a local array on each rank instead.

RussTreadon-NOAA:
Thank you @SamuelDegelia-NOAA for adding a namelist option.  

I diff'd `develop:src/enkf` and `update_mpi_readobs:src/enkf`.  As expected, multiple files differ.
```
diff -r develop/src/enkf/cmake/enkfapp_compiler_flags_Intel_Fortran.cmake pr961/src/enkf/cmake/enkfapp_compiler_flags_Intel_Fortran.cmake
diff -r develop/src/enkf/CMakeLists.txt pr961/src/enkf/CMakeLists.txt
diff -r develop/src/enkf/controlvec.f90 pr961/src/enkf/controlvec.f90
diff -r develop/src/enkf/enkf.f90 pr961/src/enkf/enkf.f90
diff -r develop/src/enkf/enkf_files.cmake pr961/src/enkf/enkf_files.cmake
diff -r develop/src/enkf/enkf_obs_sensitivity.f90 pr961/src/enkf/enkf_obs_sensitivity.f90
diff -r develop/src/enkf/gridinfo_fv3reg.f90 pr961/src/enkf/gridinfo_fv3reg.f90
diff -r develop/src/enkf/gridinfo_gfs.f90 pr961/src/enkf/gridinfo_gfs.f90
diff -r develop/src/enkf/gridinfo_nmmb.f90 pr961/src/enkf/gridinfo_nmmb.f90
diff -r develop/src/enkf/gridinfo_wrf.f90 pr961/src/enkf/gridinfo_wrf.f90
diff -r develop/src/enkf/gridio_gfs.f90 pr961/src/enkf/gridio_gfs.f90
diff -r develop/src/enkf/inflation.f90 pr961/src/enkf/inflation.f90
diff -r develop/src/enkf/innovstats.f90 pr961/src/enkf/innovstats.f90
diff -r develop/src/enkf/letkf.f90 pr961/src/enkf/letkf.f90
diff -r develop/src/enkf/loadbal.f90 pr961/src/enkf/loadbal.f90
diff -r develop/src/enkf/mpi_readobs.f90 pr961/src/enkf/mpi_readobs.f90
diff -r develop/src/enkf/params.f90 pr961/src/enkf/params.f90
diff -r develop/src/enkf/readozobs.f90 pr961/src/enkf/readozobs.f90
diff -r develop/src/enkf/readsatobs.f90 pr961/src/enkf/readsatobs.f90
diff -r develop/src/enkf/specmod.f90 pr961/src/enkf/specmod.f90
diff -r develop/src/enkf/specmod_splib.f90 pr961/src/enkf/specmod_splib.f90
```
I wonder if any of the above differences, apart from the ones in this PR, would yield more physical enkf analysis increments in RRFS tests?

[`rrfs.v1.1.0`](https://github.com/NOAA-EMC/GSI/tree/release/rrfs.v1.1.0) originated as a direct copy of [`rrfs.v1.0.0`](https://github.com/NOAA-EMC/GSI/tree/release/rrfs.v1.0.0).  `rrfs.v1.0.0` is [89 commits behind](https://github.com/NOAA-EMC/GSI/compare/release/rrfs.v1.0.0...develop) [`develop`](https://github.com/NOAA-EMC/GSI/tree/develop).  I'm always a bit leery of working on a branch so far behind the head of `develop`.

FYI, @hu5970 added a rrfs_enkf ctest (see PR #750).  It's off by default.  We should consider turning it to better monitor the impact of GSI PRs on the RRFS EnKF.

HuiLiu-NOAA:
Thanks @RussTreadon-NOAA .  
I am testing this issue using the latest GSI develop branch with RRFSv1.

SamuelDegelia-NOAA:
I just ran my test case with the `develop` branch and same job configuration. This run had similarly bad increments. 

Results using `select=2:mpiprocs=32:ompthreads=4`: 

```
ens. mean anal. increment min/max  u   -455.458129883        561.498168945    
ens. mean anal. increment min/max  v   -326.249816895        314.655395508    
ens. mean anal. increment min/max  t   -54.4604988098        305.423187256    
ens. mean anal. increment min/max  oz  -0.562584773434E-05   0.121624589156E-04
ens. mean anal. increment min/max  q  -0.389381349087       0.156813725829    
ens. mean anal. increment min/max  ps   -115.523956299        22.3902053833  
```

However, I just tried reducing the ppn and using more nodes with  `select=8:mpiprocs=8:ompthreads=4`. That resolved the problem and produced realistic increments. 

```
ens. mean anal. increment min/max  u   -17.5336685181        23.4419784546    
ens. mean anal. increment min/max  v   -27.4640083313        20.0522480011    
ens. mean anal. increment min/max  t   -3.46637058258        4.71932411194    
ens. mean anal. increment min/max  oz  -0.293705568311E-06   0.407955781156E-06
ens. mean anal. increment min/max  q  -0.497302832082E-02   0.680499663576E-02
ens. mean anal. increment min/max  ps   -2.30038237572        2.66740512848   
```

Those increments are exactly the same as what this PR produces by setting `use_shmem_window=.false.`. Reducing the ppn also solves the problem on the `release/rrfs.v1.1.0` branch. 

So it seems that giving the job more memory can resolve the increments and that this PR isn't necessarily needed. The problem might only occur for low memory jobs like we were originally using for our coarser resolution retro runs. However, I still think this problem is worth investigating, and such an option to toggle between the shared memory window and local arrays could be useful.

RussTreadon-NOAA:
Thank you @SamuelDegelia-NOAA for running these tests.  This is very useful information.

HuiLiu-NOAA:
How about the time efficiency of two options?

SamuelDegelia-NOAA:
@HuiLiu-NOAA The test case runs very fast so it is hard to get great comparisons. Here are comparisons of the two solutions using the `release/rrfs.v1.1.0` branch:

`select=2:mpiprocs=32:ompthreads=4` and `use_shmem_window=.false.` = 13.70 sec
`select=8:mpiprocs=8:ompthreads=4` and `use_shmem_window=.true.` = 13.65 sec

HuiLiu-NOAA:
Thanks @SamuelDegelia-NOAA .
This issue seems mainly a configuration problem only!

 In my opinion, it might be better to study more about the proposed change to the enkf code before adding it right now.  This would also make the enkf namelist not too crowded.

SamuelDegelia-NOAA:
Yes it seems that this problem can also be resolved through a configuration change. I tested this change in our retros and am seeing reasonable increments so far (12 cycles done).

I will go ahead and close this PR. But I will leave #960 open to continue discussing/investigating the underlying issue that causes the problem.

